PR #189, opened 2014-04-10
Add a new public method remove_watches()

rgs1:
This implementation covers up ZOOKEEPER-442. I went
with a simpler API than the one provided by the Java
implementation. If there's a need to remove individual
watchers, I think that should be provided via a separate
method (i.e.: unregister watchers) and it should be
the caller's responsability to issue a separate
remove_watches() call to remove the watch from ZK.

Signed-off-by: Raul Gutierrez S rgs@itevenworks.net

### DIFF
--- a/kazoo/client.py
+++ b/kazoo/client.py
@@ -34,6 +34,7 @@ from kazoo.protocol.serialization import (
     Exists,
     GetChildren,
     GetChildren2,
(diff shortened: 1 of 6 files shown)


bbangert:
Also, the bare print statement causes a Python 3 syntax error.

bbangert:

PR #206, opened 2014-05-21
Add support for non-ephemeral leases in the semaphore recipie

jacksontj:
This is useful when the task you are tring to limit concurrency (such as a deployment) when a failure should retain a lock.


bbangert:

PR #214, opened 2014-06-11
Make sure Semaphore uses get_lease inside of lock

vishvananda:
In non-blocking mode it was possible for the semaphore to acquire
the lease without being inside a lock, so mor than max_leases couild
acquire the semaphore simultaneously. This fixes the issue by ensuring
that _get_lease is always called inside the lock.


bbangert:

Issue #222, opened 2014-07-17
Transcation reports true even though it was stopped


harlowja:
It appears that when a transaction is used and part of that transaction is a check() command that when that check command fails the transaction object still reports that it has committed (even though it really has not).

Example that can trigger this.

```
from kazoo import client
from kazoo import exceptions as k_exceptions

c = client.KazooClient()
c.start()
try:
    c.create("/tmp222")
except k_exceptions.NodeExistsError:
    pass

c_data, c_stat = c.get("/tmp222")
txn = c.transaction()
txn.check("/tmp222", version=c_stat.version + 1)
txn.create("/bbbbbb")

print txn.commit()
print txn.committed
c.get("/bbbbbb")
```

This outputs:

```
[BadVersionError((), {}), RuntimeInconsistency((), {})]
True
Traceback (most recent call last):
  File "test2.py", line 21, in <module>
    c.get("/bbbbbb")
  File "/home/harlowja/Dev/taskflow/.tox/py27/local/lib/python2.7/site-packages/kazoo/client.py", line 978, in get
    return self.get_async(path, watch).get()
  File "/home/harlowja/Dev/taskflow/.tox/py27/local/lib/python2.7/site-packages/kazoo/handlers/threading.py", line 102, in get
    raise self._exception
kazoo.exceptions.NoNodeError: ((), {})
```

I would expect `committed` to show `False` when a check() fails?

This is using zookeeper 3.4.5 (if that matters).

PR #224, opened 2014-07-19
Rawlink a callback on transaction completion

harlowja:
Instead of assuming that a transaction will
correctly complete before it actually does we
should be smarter about this and inspect the
result to determine if it actually did and only
if it did should we set the committed flag.

Fixes issue #222


bbangert:

PR #225, opened 2014-07-25
Add a checked commit transaction method

harlowja:
Instead of just having a commit method that leaves
the majority of the failure handling to the caller it
is nice to have a checked commit method that does most
of this work onbehalf of the caller (and throws an
exception when errors are detected).


bbangert:

PR #240, opened 2014-08-26
Always use async=True for SetPartitioner._child_watching()

emulbreh:


bbangert:
This breaks Zookeeper 3.3.6 compatibility according to Travis.

bbangert:

Issue #227, opened 2014-08-07
Cumbersome implementation for non-blocking Locks


stevemer:
Preface and disclaimer: I am a newer coder than most of y'all, so be gentle!

---

Opinion: This is not very clean and simple:

```
zk = KazooClient()
lock = zk.Lock("/lockpath", "my-identifier")
lock.acquire(blocking=False)
if lock.is_acquired:
  # do something
```

---

---

Opinion: This is cleaner and simpler:

```
zk = KazooClient()
with zk.Lock("/lockpath", "my-identifier", default_locking=False) as lock:
    if lock.is_acquired:
        # do something
```

And would require simply adding default_blocking=True to **init_** (+ a few more lines).

---
## Proposal:

Could we add a parameter to specify blocking behavior from **init**? Wouldn't change existing behavior, would (in my opinion) allow for cleaner coding, and I'd be happy to write it.

---

As a side rant, if pep-0377 had passed (:( ), this would be the cleanest and simplest I think?

```
zk = KazooClient()
with zk.Lock("/lockpath", "my-identifier", default_locking=False) as lock:
    # throw SkipException() out of __enter__ and catch into __exit__
    # do something only gets hit if we got the lock
```

---

PR #242, opened 2014-09-04
(#227) allow for usage of non-blocking locks inside with statements

stevemer:
Introduce an optional argument to **init** allowing for specification of blocking behavior at instantiation while maintaining backwards compatibility.

This allows for use of non-blocking Locks inside with statements.

---

```
zk = KazooClient()
with zk.Lock("/lockpath", "my-identifier", blocking=False) as lock:
    if lock.is_acquired:
        # do something
```

---

### DIFF
--- a/kazoo/recipe/lock.py
+++ b/kazoo/recipe/lock.py
@@ -40,13 +40,23 @@ class Lock(object):
         with lock:  # blocks waiting for lock acquisition
             # do something with the lock
 
+    Example usage for a non-blocking lock implementation:
+
+    .. code-block:: python
+
+        zk = KazooClient()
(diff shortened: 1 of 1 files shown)


kazoo/recipe/lock.py
```diff
@@ -106,6 +120,7 @@ def acquire(self, blocking=True, timeout=None):
         .. versionadded:: 1.1
             The timeout option.
         """
+        blocking = self.blocking if blocking else blocking
```
harlowja: I agree; I'd rather have this blocking take precedence over the instance one.

Otherwise it just seems confusing; maybe have ``blocking is None`` then use the instance one?

bbangert:

PR #257, opened 2014-11-28
PatientChildrenWatch can lose watchers in case of connection suspend

kovalevfm:
Hi.
If one zookeeper server suspends in cluster PatientChildrenWatch can lose watchers.
It lead to incorrect SetPartitioner work.
I've added session_watcher and checking that children same as before suspend.
It starts work after _inner_start finished


bbangert:
Please re-open to rebase and update as needed. This looks good, but it needs tests or I don't know if it's doing anything useful.

PR #258, opened 2014-12-02
Proper check if we hold the lock in locking queue

nekto0n:
There's a misprint - instead of checking if we're holding the lock, we just checking method like `bool(self.holds_lock)`.


nekto0n:
Can someone restart travis build? Seems like it failed for some internal reason.

bbangert:
If a test breaks from this, I'll fix it, cause this is a clear code bug.

PR #268, opened 2015-01-28
Fixed endless loop problem in connection handler

max0d41:
The problem was that `select` could return more than one element, but only the first one was checked which could result in some situations in an endless loop.


paladinize:
@max0d41 thanks for the fix! Any chance this can be merged by someone in the near future?

bbangert:
I see no reason this shouldn't work, since there's no guarantee self._socket must be the first element.

max0d41:
I thought this was the issue causing an endless loop, but finally I found our problem. Our app was blocking too long when HDD IO was very high. I moved the affected code to gevent ThreadPool threads and now all the connection problems with zookeeper are gone.

PR #271, opened 2015-02-10
Fix handling KazooState.SUSPENDED in SetPartitioner

KonishchevDmitry:
We must do a repartition at every connection state change because all watchers are lost when connection gets ``SUSPENDED`` state.

I've also added ``state_change_event`` attribute to be able to respond more quickly to partitioner state changes.


bbangert:
Looks good, any chance of a test?

KonishchevDmitry:
Yes, I'm on it.

KonishchevDmitry:
@bbangert, I've added the tests. Please look on it. One build has failed, but it seems that it's due to a some temporary problem on a build server.

bbangert:
I kicked travis, tests now pass. Looks like code coverage went up a little as well.

kazoo/tests/test_partitioner.py
```diff
@@ -8,85 +9,140 @@
...
-            self.path, set=(1, 2, 3), time_boundary=0.2)
-        partitioner.wait_for_acquire(14)
-        eq_(partitioner.state, PartitionState.ACQUIRED)
-        eq_(list(partitioner), [1, 2, 3])
-        partitioner.finish()
+        self.__create_partitioner(size=3)
+        self.__wait_for_acquire()
+        self.__assert_state(PartitionState.ACQUIRED)
+        self.__assert_partitions([0, 1, 2])
+        self.__finish()
```
bbangert: What's the rationale for using private class methods vs. just normal private methods?

harlowja:
Only one question/comment/nit.

Whats up with all the ``self.__`` usage in tests; haven't seen that used much (especially in tests where its not that much of a concern to really hide members/functions...)?

KonishchevDmitry:
When I write some class I always make all members private until it become needed by derived classes or class users for the following reasons:
* When I expose some API that may be used by someone else I take a burden of supporting this API. So I believe that it's better to expose any member only if it may be needed by derived classes or class users.
* If I want to refactor some code, I always feel free to modify any ``self.__*`` members without searching for their usage in other modules because ``__`` guarantees that this member is used only in this class.
* A can name my methods using short names like ``__wait()`` and ``__finish()`` without any chance to accidentally override other member in one of base classes. For example there is a chance that in the future ``KazooTestCase`` will get a ``wait()`` or ``_wait()`` method which will wait until the connection will be established. By naming my method as ``__wait()`` I protect myself from possible colliding of two methods with totally different meaning.

But of course it's not principal - it's just my coding style and if you wish I'll replace all my ``self.__member`` with ``self._member``.

KonishchevDmitry:
Is there anything I should do to make the PR be merged?

bbangert:
Nope, looks good to me.

PR #273, opened 2015-02-12
Add an eventlet based handler

harlowja:
Integrate eventlet *handler* support into kazoo so that along with [gevent,
threading] handlers there can now be a dedicated eventlet handler for
projects that need to (or want to) use eventlet (such as those working
in the openstack community). The ``requirements_eventlet.txt`` file lists
the optional eventlet requirement(s) that needs to be satisfied when this
new handler is used.


bbangert:
Doesn't this need to modify travis.yml to install eventlet so that travis runs its tests?

harlowja:
Thought I got that, might have not; checking.

bbangert:
I think you did, travis says it ran eventlet tests.

bbangert:
r+

PR #275, opened 2015-02-13
Tweak some of the import(s)

harlowja:
Avoid late importing modules that we can just import at the
normal top of file time (late importing doesn't seem to add
any benefit and IMHO makes the code harder to understand).

Also order more of the imports to follow the standards where
they are typically ordered alphabetically (with stdlib modules
being before other modules and the library itself being last).


bbangert:
r+, looks fine

PR #276, opened 2015-02-13
Use six for *some of* py3/py2 compat tweaks

harlowja:
Instead of doing customizations for py2/py3 just use
the six module (which is all this module does) to achieve
the same effect but in a more maintained and standard
way.

This uses the six module to do the following:

* Figure out what the best urllib parse is at.
* Figure out and use the byte/string types that six
  locates in the client (and adjust the type checks
  and exceptions raised to use it); also improves
  upon the type errors raised to hopefully be more
  meaningful.

There are still a few leftover but hopefully those can
be adjusted in the future (and in another patch series).


bbangert:
r+, nice cleanup

PR #279, opened 2015-02-13
fix attribute error when deleting node if self.node was not yet assigned

trbs:
The lock recipe can fail with attribute error when connection error happens before `self.node = node` is called.

Fixed by explicitly defining `self.node = None` in `__init___` and checking it before deleting the node when lock could not be acquired.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -29,6 +29,10 @@ Bug Handling
 
 - #261: Creating a sequential znode under / doesn't work.
 
+- #279: Fixed AttributeError 'Lock' object has no attribute 'node'. Caused
+        by the missing definition of self.node in __init__.
+
+
 Documentation
 *************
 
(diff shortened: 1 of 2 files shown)


bbangert:
I'm actually not sure a test can reliably cause this, but it definitely deserves a changelog entry.

trbs:
We hit the problem because we have a lock that does not retry when an error happens trying to acquire the lock. When we restarted one of our ZK nodes we got this ```AttributeError``` on the ```Lock``` object. I assume this is also why nobody else hit and reported the issue. As normally one will try and get another ZK node to try and acquire the lock.

If kazoo has some test infrastructure to simulate a ZK failure after N operations then we might be able to test this reliably by KazooRetry(max_tries=0) and have the mock ZK node fail after the first ZK operation (ensurePath?) in ```_inner_acquire```.

Reading the recipe code I was expecting this to be a programming error that slipped into the code. I cannot find a reason why ```self.node``` is not explicitly declared in ```__init__``` while the other major attributes of the ```Lock``` are.

trbs:
changelog entry added

bbangert:

Issue #274, opened 2015-02-13
Periodically able to cause `server_version` to fail


harlowja:
It appears like the following can be triggered (for unknown reasons) when calling into client.server_version()

```
File "/usr/lib/python2.7/dist-packages/kazoo/client.py", line 642, in server_version
string = ENVI_VERSION.match(data).group(1)
AttributeError: 'NoneType' object has no attribute 'group' 
```

This is happening periodically on kazoo 2.0 (periodically, and not all the time for rackspace who is seeing this intermittently); so this could be a zookeeper (happens on 3.4.5) bug or a kazoo one (or a mix of both).

harlowja:
More data is being collected to see exactly what the server_version 'envi' command is returning.

harlowja:
So talking with folks it seems this is known to be somewhat flakey due to the internals of zookeeper and the '4' letter word commands being known to not be very reliable:

https://issues.apache.org/jira/browse/ZOOKEEPER-1346 will hopefully make this better in the future.

rgs1:
Yup. If you try:

```python
client.command('dump')
```
you'll see you get partial data some times too. That's because ZK closes the socket immediately after writing/flushing, which means you could still lose some.

bbangert:
Should it maybe retry the operation and return None if there's an error?

harlowja:
IMHO the hard part to determine is the 'error'; it could likely retry it, but what if it gets an environment like the following:

```
Environment:
zookeeper.version=3.<eom>
```

Or it could be truncated at:
```
Environment:
zookeeper.version=3.4.2
blah=blah
blah2=<eom>
```
The hard part is the first one (where the truncation happens right in the version number, the lack of a valid string can be easily retried, but a partial version string is the tough one, are they always 3 digits, if so, then it can likely be retried if its not 3 digits; otherwise its sorta hard to tell the difference between a partial version and one that isn't partial).

harlowja:
This has been addressed (for now, until better zookeeper apis come out).

PR #280, opened 2015-02-15
Add retries to server_version() fetch

harlowja:
Due to the apparent flakiness of the four-letter commands
we need to add a set of retries to the fetching of the
server version to handle getting partial or truncated values
back from it.

This adds a default of 4 tries (1 + 3 retries) to the server
version code so that if a partial value is returned (or no
server version at all) that the code will try again, if none
can be fetched in 4 tries; then none is returned to signify
this failure.

Fixes issue #274

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -32,6 +32,9 @@ Bug Handling
 
 - #261: Creating a sequential znode under / doesn't work.
 
+- #274: Add server_version() retries (by default 4 attempts will be made) to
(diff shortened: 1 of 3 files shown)


jimbobhickville:
+1 This would be very helpful for us over here.

bbangert:
Should be testable by having a mock return an invalid reply the first call.

harlowja:
Will do.

PR #283, opened 2015-02-20
Fix a race condition in SetPartitioner when party changes during handling of lock acquisition

KonishchevDmitry:
This pull request is made on top of #271. I've noticed that there is a race condition due to which we might get a deadlock or invalid partition result.

In https://github.com/KonishchevDmitry/kazoo/commit/808aa9207e0788542b8af7bf5e347904644a172e I've wrote two tests for race conditions that current version of SetPartitioner doesn't pass and than fixed the error in https://github.com/KonishchevDmitry/kazoo/commit/7fb411b92a317c5edd0225aab6f3a24e5516b969.


KonishchevDmitry:
The failed tests seems to be not my fault.

bbangert:
Kicked a new build off.

bbangert:
r+, this looks good to me pending the test.

PR #284, opened 2015-02-26
Add support for communicating with zookeeper with SASL

### LINKED ISSUE #130: Add Kerberos authentication support
As of version 3.4, Zookeeper allows authentication via Kerberos, but at the moment only the Java client supports it. Are there plans to add this to Kazoo in the near future?

erickt:
This adds initial support for communicating with Zookeeper using SASL (by way of [python-sasl](https://github.com/toddlipcon/python-sasl)) for authentication. It however is not doing any encryption between the server and the client. To connect with SASL, you must make sure you have `kinit`-ed a kerberos session, and then pass a `sasl_server_principal` in the `KazooClient` constructor, as in:

``` python
host = ...
port = ...
zk = kazoo.client.KazooClient(hosts=host + ':' + port, read_only=True, sasl_server_principal="zookeeper")
```

This will need some tests before it lands, but I wanted to file this here first to make sure the approach is correct.

Closes #130.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -24,6 +24,8 @@ Features
 - Add a NonBlockingLease recipe.  The recipe allows e.g. cron jobs scheduled
   on multiple machines to ensure that at most N instances will run a particular
   job, with lease timeout for graceful handover in case of node failures.
+- Add ``sasl_server_principal`` to ``kazoo.client.Client``, which enables kazoo
+	to authenticate against a kerberized Zookeeper.
 
 
 Bug Handling
--- a/kazoo/client.py
+++ b/kazoo/client.py
@@ -105,8 +105,8 @@ class KazooClient(object):
     """
     def __init__(self, hosts='127.0.0.1:2181',
                  timeout=10.0, client_id=None, handler=None,
-                 default_acl=None, auth_data=None, read_only=None,
-                 randomize_hosts=True, connection_retry=None,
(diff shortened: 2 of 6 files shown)


erickt:
@bbangert: I've added a note to the changelog and rebased it on HEAD. Tests will take a little longer in order to figure out how to figure out how to setup a test kerberized environment.

jimbobhickville:
@erickt you can configure zookeeper with a digest-based SASL so you don't have to bring in kerberos to the equation.  Is there any update on this patch otherwise?  We need SASL support here, so if you've abandoned it, we're happy to pick it back up and finish it off.

erickt:
@jimbobhickville: A belated hello! I don't have much time to finish up this patch, but we have been running it in production in https://github.com/cloudera/hue, so it should be pretty robust.

kyprizel:
Any chance to see it in upsteam?

bbangert:
Please re-open to rebase and update as needed. @jimbobhickville, if you'd like to update this with a basic test (even if it has to use mocks), I'd be happy to merge.

superflav:
@jimbobhickville, did your team make any further progress on this? We've got a Python app using Kazoo that needs to communicate w/a Kerberized Zookeeper, and this looks like it may do the trick.

superflav:
Thanks for the speedy reply, @jimbobhickville . We'll have to take this over the line if it does what we need it to.

PR #287, opened 2015-03-03
Support SASL/GSSAPI authentication

mkomitee:
To trigger SASL/GSSAPI authentication, add an auth_data tuple of `("gssapi", None)` to the `KazooClient`. It uses the users current Kerberos credential cache to authenticate.

This was developed against 1.3.1, and so I fully expect it to *not* merge against the current HEAD of kazoo.

Furthermore it was developed against an older version of PyGSSAPI which included a bug. I fixed that bug in my own copy of PyGSSAPI, but when I went to submit the fix upstream, PyGSSAPI had been significantly modified and it was unbuildable at the time.

I'm only submitting this PR to serve as a reference for #130 and #284.

### DIFF
--- a/kazoo/client.py
+++ b/kazoo/client.py
@@ -36,6 +36,7 @@ from kazoo.protocol.serialization import (
     GetACL,
     SetACL,
     GetData,
+    SASLAuth,
     SetData,
     Sync,
     Transaction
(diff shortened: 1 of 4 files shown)


bbangert:
Ok, so I'm closing this then since there's no intention to update it to cleanly merge, tests, etc. It also add's the requirement to kazoo to install PyGSSAPI, which many people don't need so it shouldn't be there.

PR #289, opened 2015-03-13
Use the helper python2atexit for atexit routines

harlowja:
Use the right ``atexit`` routines (that work better in py2/py3).


thesquelched:
I can confirm that this fixes the issue that I am experiencing with hanging `SequentialEventletHandler` objects.

PR #290, opened 2015-03-13
Allow disabling the atexist register/unregister

harlowja:


bbangert:

PR #293, opened 2015-03-17
Have the lock recipe no longer be sorta re-entrant

### LINKED ISSUE #291: Lock or RLock ?
Hi, 

I'm a bit confused about the kind of lock that kazoo implements, it's called Lock, but the documentation said it's a reentrant lock, but when I test it the release() don't look reentrant:

  lock.acquire() # return True
  lock.acquire() # return True
  lock.release() # release the lock
  lock.release() # should release the lock, but that have been already done by the previous call, code executed between both release() is no more protected by the lock...

Regards,

harlowja:
Instead of having acquire be re-entrant but release remove
the lock immediately we should just not be re-entrant at
all since the current implemetnation is mostly broken in
the first place (especially when the inner release will destroy
other acquires...).

Fixes #291

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -23,6 +23,13 @@ Features
 Bug Handling
 ************
 
+- #291: Kazoo lock recipe was only partially re-entrant in that multiple
+  calls to `acquire` would obtain the the lock but the first call to `release`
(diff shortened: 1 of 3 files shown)


bbangert:
r+, waiting on travis for merge

PR #295, opened 2015-03-18
Use the spawn method when creating thread handler workers

harlowja:
Instead of creating daemon threads in two places, both
using the same code, just have the `_create_thread_worker`
method call into the spawn method to delegate to it to create
a daemon thread.


bbangert:
r+

PR #296, opened 2015-03-18
Add tox.ini for using tox for testing

harlowja:
Refactor parts of the makefile and testing to
support the currently standard tox command to test
various environments.

This also moves most of the zookeeper setup and
teardown logic out of the makefile and into a specific
script file that just does this setup/teardown.


bbangert:
r+

PR #297, opened 2015-03-19
Have the lock reentrancy behave like threading.lock

harlowja:
Instead of having it raise a runtime error have the calling
thread block until the lock has been released (typically
by some other thread). This more closely matches the existing
and (assumingly) expected behavior of locks.

### DIFF
--- a/kazoo/tests/test_lock.py
+++ b/kazoo/tests/test_lock.py
@@ -1,4 +1,5 @@
 import uuid
+import time
 import threading
 
 from nose.tools import eq_, ok_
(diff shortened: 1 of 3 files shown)


kazoo/tests/test_lock.py
```diff
@@ -244,11 +245,33 @@ def test_lock_no_double_calls(self):
         lock1 = self.client.Lock(self.lockpath, "one")
         lock1.acquire()
         self.assertTrue(lock1.is_acquired)
-        self.assertRaises(RuntimeError, lock1.acquire)
+        self.assertFalse(lock1.acquire(timeout=0.5))
         self.assertTrue(lock1.is_acquired)
         lock1.release()
         self.assertFalse(lock1.is_acquired)
 
+    def test_lock_many_threads(self):
```
bbangert: How exactly is this testing that the other threads are blocking?

Here's how I would test the blocking behavior. Spin out the threads, use a threading.Condition in each one. Check that they're blocked (as they can't append to acquires), and as you release each Condition in turn (allowing the acquire of the next to complete), check that the acquires grows by 1 each time. The test as-is would pass if the lock was re-entrant since they'd all acquire it and append to the list.

harlowja: Agreed; let me make this better.

PR #305, opened 2015-04-03
[RFC] Support SetWatches and watch restoration

### LINKED ISSUE #218: Support for SetWatches command, as well as watch restoration on ZK session reconnect
I've hacked in SetWatches request into a local version of Kazoo for testing.

I'm not sure if this would be a feature that's relatively useful for people to expose as an API, but I think it could be useful in the case where a connection is dropped and removed from the server before a session expires, as can be the case in client restart.

Example: Client exits abnormally. The original connection is loss, the ZooKeeper server removes the connection, which removes all watches associated with that connection (as the ServerCnxn object itself is the watcher on the server side).

User had persisted KazooClient.client_id, as well as information about paths that were being watched, and KazooClient.last_zxid.

Upon reconnect, user supplies that information back to KazooClient.
User reconstructs watcher(s).

KazooClient starts, reconnecting with the previous session ID, restoring the previous session, and immediately sets watches.

If anything has changed since the relativeZxid, those events are immediately triggered (which is why client-side watch handlers should be to be set before this request).

Here is an example of the SetWatches for serialization.py:

```
class SetWatches(namedtuple('SetWatches', 'relativeZxid dataWatches existsWatches childWatches')):
    type = 101

    def serialize(self):
        b = bytearray()
        b.extend(long_struct.pack(self.relativeZxid))
        self.svector(b, self.dataWatches)
        self.svector(b, self.existsWatches)
        self.svector(b, self.childWatches)
        return b

    def svector(self, b, arr):
        b.extend(int_struct.pack(len(arr)))
        for elem in arr:
            b.extend(write_string(elem))

    # Not sure what we need for deserialization here
```

rgs1:
Basic plumbing to support resetting watches upon recovering a session (as done by the Java implementation).

Given this breaks the current behaviour, it should be made optional. I'll circle back with that added in. Comments welcomed! 

Fixes #218 

Signed-off-by: Raul Gutierrez S rgs@itevenworks.net

### DIFF
--- a/kazoo/client.py
+++ b/kazoo/client.py
@@ -453,7 +453,8 @@ class KazooClient(object):
             self._live.clear()
             self._notify_pending(state)
             self._make_state_change(KazooState.SUSPENDED)
-            self._reset_watchers()
+            if state != KeeperState.CONNECTING:
+                self._reset_watchers()
 
     def _notify_pending(self, state):
         """Used to clear a pending response queue and request queue
--- a/kazoo/protocol/connection.py
+++ b/kazoo/protocol/connection.py
@@ -26,6 +26,7 @@ from kazoo.protocol.serialization import (
     Ping,
     PingInstance,
     ReplyHeader,
+    SetWatches,
     Transaction,
     Watch,
     int_struct
@@ -59,6 +60,7 @@ CHILD_EVENT = 4
 WATCH_XID = -1
 PING_XID = -2
 AUTH_XID = -4
+SET_WATCHES_XID = -8
 
 CLOSE_RESPONSE = Close.type
 
@@ -406,6 +408,8 @@ class ConnectionHandler(object):
                 async_object.set(True)
         elif header.xid == WATCH_XID:
             self._read_watch_event(buffer, offset)
+        elif header.xid == SET_WATCHES_XID:
+            self.logger.log(BLATHER, 'Received SetWatches reply')
         else:
             self.logger.log(BLATHER, 'Reading for header %r', header)
 
@@ -438,6 +442,8 @@ class ConnectionHandler(object):
         # Special case for auth packets
         if request.type == Auth.type:
             xid = AUTH_XID
+        elif request.type == SetWatches.type:
+            xid = SET_WATCHES_XID
         else:
             self._xid += 1
             xid = self._xid
@@ -588,6 +594,10 @@ class ConnectionHandler(object):
                           client._session_id or 0, client._session_passwd,
                           client.read_only)
 
+        # save the client's last_zxid before it gets overwritten by the server's.
+        # we'll need this to reset watches via SetWatches further below.
+        last_zxid = client.last_zxid
+
         connect_result, zxid = self._invoke(
             client._session_timeout / 1000.0, connect)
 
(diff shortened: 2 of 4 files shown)


tonyseek:
It seems the connection will be broken if we are using data watcher and child watcher in the same client. The following test could reproduce this:

``` diff
diff --git a/kazoo/tests/test_client.py b/kazoo/tests/test_client.py
index af8ecf5..e082232 100644
--- a/kazoo/tests/test_client.py
+++ b/kazoo/tests/test_client.py
@@ -965,16 +965,20 @@ class TestClient(KazooTestCase):
 
     def test_set_watches_on_reconnect(self):
         client = self.client
-        watch_event = client.handler.event_object()
+        data_watch_event = client.handler.event_object()
+        child_watch_event = client.handler.event_object()
 
         client.create("/tacos")
 
         # set the watch
-        def w(we):
-            eq_(we.path, "/tacos")
-            watch_event.set()
+        def make_watch(watch_event):
+            def w(we):
+                eq_(we.path, "/tacos")
+                watch_event.set()
+            return w
 
-        client.get_children("/tacos", watch=w)
+        client.get("/tacos", watch=make_watch(data_watch_event))
+        client.get_children("/tacos", watch=make_watch(child_watch_event))
 
         # force a reconnect
         states = []
@@ -986,19 +990,26 @@ class TestClient(KazooTestCase):
                 states.append(state)
                 rc.set()
 
-        client._connection._socket.shutdown(socket.SHUT_RDWR)
+        # client._connection._socket.shutdown(socket.SHUT_RDWR)
+        self.lose_connection(threading.Event)
 
         rc.wait(10)
         eq_(states, [KazooState.CONNECTED])
 
         # watches should still be there
-        self.assertTrue(len(client._child_watchers) == 1)
+        self.assertEqual(len(client._data_watchers), 1)
+        self.assertEqual(len(client._child_watchers), 1)
 
         # ... and they should fire
         client.create("/tacos/hello_", b"", ephemeral=True, sequence=True)
 
-        watch_event.wait(1)
-        self.assertTrue(watch_event.is_set())
+        child_watch_event.wait(1)
+        self.assertTrue(child_watch_event.is_set())
+
+        client.set("/tacos", b"")
+
+        data_watch_event.wait(1)
+        self.assertTrue(data_watch_event.is_set())
 
 
 dummy_dict = {
```

This may be the root cause which breaks the test of #398.

I am trying to solve this.

bbangert:
@tonyseek if you revert this PR, does the issue go away? I think I've seen a reported case where it had a XIDS mismatch also looking for something like this.

tonyseek:
@bbangert I believe this is a race condition bug which introduced by the SetWatch and `_invoke` implementation. The received response header is pointed to previous watch events instead of the SetWatch request. I am still tracking on this issue, and it will go away if we revert this PR.

bbangert:
Ok, I'm inclined to revert this then, as this wasn't your PR I don't expect you to fix it. Thank you for diving into this to track down how it's blocking the other PR.

bbangert:
@tonyseek I've reverted this PR and merged to master. The TreeCache PR might 'just work' now.

brandonDu:
I do not see the change in the connection.py, when can we use the library for the change.

bbangert:
@brandonDu the revert for this fix is in the latest master, it hasn't been released as a new revision yet.

brandonDu:
@bbangert , I just test the case, what is worse is that the watch can not work.
I think most of the code u commit  works well ,except the raise RuntimeError.

tonyseek:
@brandonDu The recipes of Kazoo like `DataWatch` and `ChildrenWatch` work well without the `SetWatch` support. Because they have implemented the session watcher correctly. I suggest using a tested recipe like them instead of processing raw watch events by yourself.

The RuntimeError is critical and worse than lack of `SetWatch`. It is caused by packets which out of order and it breaks connection during reconnecting.

rajath26:
Was setWatcher and watch restoration ever brought back in the later versions of kazoo?

rgs1:
Ok, sorry for joining the party a few years late but the fix is really pretty easy: 

given it's fine to receive either an xid -8 (SetWatches) or xid -1 (WatchEvent) in the response when setting SetWatches (since a watch can fire immediately after being set), we just need to change `_invoke()` to accept a list of valid xids instead of just one. 

Happy to send a PR to unrevert and handle this scenario.

StephenSorriaux:
Hello,

Thank you for looking into this. We would be very happy to review a PR for this.

PR #306, opened 2015-04-06
Implement read-write (shared) lock from ZooKeeper recipe.

cpatulea:
Reference: http://zookeeper.apache.org/doc/trunk/recipes.html#Shared+Locks

This is a writer-preference shared lock.

Most of the lock protocol is the same as for an exclusive lock, except that it needs to be parameterized on:
1) Node name - lock "kind" - a write lock (**lock**) or read lock (**rlock**).
2) Which node names block which other node names (read locks block only writers, writers block both).
3) Finding the predecessor node according to the rule in #2.

These changes are integrated to the main Lock class. The changes are backward compatible with existing users of Lock. Users should normally never pass the Lock node_name and exclude_names kwargs directly, they should use WLock and RLock.

### DIFF
--- a/kazoo/recipe/lock.py
+++ b/kazoo/recipe/lock.py
@@ -217,7 +217,14 @@ class Lock(object):
     def _get_sorted_children(self):
         children = self.client.get_children(self.path)
 
-        # can't just sort directly: the node names are prefixed by uuids
+        # Node names are prefixed by uuids: strip the prefix first, which may
+        # be one of multiple values in case of a read-write lock, and return
+        # only the sequence number (as a string since it is padded and will sort
+        # correctly anyway).
+        #
+        # In some cases, the lock path may contain nodes with other prefixes
+        # (eg. in case of a lease), just sort them last ('~' sorts after all
+        # ASCII digits).
         def _seq(c):
             for name in ["__lock__", "__rlock__"]:
                 idx = c.find(name)
--- a/kazoo/recipe/lock.py
+++ b/kazoo/recipe/lock.py
@@ -43,11 +43,18 @@ class Lock(object):
     Note: This lock is not *re-entrant*. Repeated calls after already
     acquired will raise a ``RuntimeError``.
 
+    This is an exclusive lock. For a read/write lock, see :method:`WLock` and
+    :method:`RLock`.
+
     """
 
-    def __init__(self, client, path, identifier=None, node_name=None,
+    def __init__(self, client, path, identifier=None, node_name="__lock__",
                  exclude_names=None):
         """Create a Kazoo lock.
+
+        node_name and exclude_names are typically only used internally to
+        implement read/write locks. They should be left unset for exclusive
+        locks.
 
         :param client: A :class:`~kazoo.client.KazooClient` instance.
         :param path: The lock path to use.
@@ -71,8 +78,6 @@ class Lock(object):
 
         self.wake_event = client.handler.event_object()
 
-        if node_name is None:
-            node_name = "__lock__"
         self.node_name = node_name
 
         if exclude_names is None:
@@ -299,7 +304,7 @@ def WLock(*args, **kwargs):
     .. code-block:: python
 
         zk = KazooClient()
-        lock = WLock("/lockpath", "my-identifier")
+        lock = WLock(zk, "/lockpath", "my-identifier")
         with lock:  # blocks waiting for any outstanding readers and writers
             # do something with the lock
 
@@ -324,7 +329,7 @@ def RLock(*args, **kwargs):
     .. code-block:: python
 
         zk = KazooClient()
-        lock = RLock("/lockpath", "my-identifier")
+        lock = RLock(zk, "/lockpath", "my-identifier")
         with lock:  # blocks waiting for any outstanding writers
             # do something with the lock
 
--- a/kazoo/recipe/lock.py
+++ b/kazoo/recipe/lock.py
@@ -218,7 +218,8 @@ class Lock(object):
                 idx = c.find(name)
                 if idx != -1:
                     return c[idx + len(name):]
-            raise ValueError("Unknown node type: %s" % c)
+            # Sort unknown node names eg. "lease_holder" last.
+            return '~'
         children.sort(key=_seq)
         return children
 
--- a/kazoo/recipe/lock.py
+++ b/kazoo/recipe/lock.py
@@ -217,7 +217,7 @@ class Lock(object):
     def _get_sorted_children(self):
         children = self.client.get_children(self.path)
 
-        # Node names are prefixed by uuids: strip the prefix first, which may
+        # Node names are prefixed by a type: strip the prefix first, which may
         # be one of multiple values in case of a read-write lock, and return
         # only the sequence number (as a string since it is padded and will sort
         # correctly anyway).
--- a/kazoo/recipe/lock.py
+++ b/kazoo/recipe/lock.py
@@ -44,9 +44,9 @@ class Lock(object):
     acquired will raise a ``RuntimeError``.
 
     """
-    _NODE_NAME = '__lock__'
 
-    def __init__(self, client, path, identifier=None):
+    def __init__(self, client, path, identifier=None, node_name=None,
+                 exclude_names=None):
         """Create a Kazoo lock.
 
         :param client: A :class:`~kazoo.client.KazooClient` instance.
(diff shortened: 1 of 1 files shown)


kazoo/recipe/lock.py
```diff
@@ -54,7 +54,13 @@ def __init__(self, client, path, identifier=None):
...
                            can be useful for querying to see who the
                            current lock contenders are.
-
+        :param node_name: Node name, after the contender UUID, before the
+                          sequence number. Involved in read/write locks. For a
+                          normal (exclusive) lock, leave unset.
+        :param exclude_names: Node names which exclude this contender when
+                              present at a lower sequence number. Involved in
+                              read/write locks. For a normal (exclusive) lock,
+                              leave unset.
```
kamalmarhubi: I'd consider moving the comment that these are for read/write locks to the front, so that uninterested people know they can ignore them.

kazoo/recipe/lock.py
```diff
@@ -65,11 +71,19 @@ def __init__(self, client, path, identifier=None):
 
         self.wake_event = client.handler.event_object()
 
+        if node_name is None:
+            node_name = "__lock__"
+        self.node_name = node_name
```
kamalmarhubi: strings are immutable, so you could safely put the default in the arglist

kazoo/recipe/lock.py
```diff
@@ -44,9 +44,9 @@ class Lock(object):
     acquired will raise a ``RuntimeError``.
 
     """
-    _NODE_NAME = '__lock__'
```
kamalmarhubi: maybe leave this in as _DEFAULT_NODE_NAME?

kazoo/recipe/lock.py
```diff
@@ -44,9 +44,9 @@ class Lock(object):
     acquired will raise a ``RuntimeError``.
 
     """
```
kamalmarhubi: Do you want to add mention of the reader/writer lock in the class docstring?

kazoo/recipe/lock.py
```diff
@@ -267,6 +291,56 @@ def __exit__(self, exc_type, exc_value, traceback):
...
 
+def WLock(*args, **kwargs):
+    """Kazoo read-write lock (writer side).
+
+    Example usage:
+
+    .. code-block:: python
+
+        zk = KazooClient()
+        lock = WLock("/lockpath", "my-identifier")
```
kamalmarhubi: Obviously I don't know the code base, but the creation and non-use of `zk` looks fishy to me. Is there an implicit global client?

kamalmarhubi:
High level comment: it looks as though nodes following the read/write lock protocol could interact with nodes following the exclusive lock protocol at the same node name. This could lead to safety issues since the exclusive lock protocol doesn't exclude `__rlock__`: a node following the exclusive protocol would happily take the lock even if there is a reader, and go on its merry way.

I may be wrong, in a morning code review rush. :-)

kazoo/recipe/lock.py
cpatulea: Done.

kazoo/recipe/lock.py
cpatulea: Done.

kazoo/recipe/lock.py
cpatulea: Eh. This also gets used in WLock and RLock, so module scope would be a more appropriate place.

But I'm not sure there's a point in defining symbolic constants for these anyway. They're already just arbitrary tokens, they're readily searchable, and they're unlikely to ever change, so I figure 'explicit' in this case would aid readability.

kazoo/recipe/lock.py
cpatulea: Done.

kazoo/recipe/lock.py
cpatulea: Thanks, done.

harlowja: That shouldn't be the case (no global), probably just the docs are not right here.

kazoo/recipe/lock.py
```diff
@@ -267,6 +291,56 @@ def __exit__(self, exc_type, exc_value, traceback):
...
+
+
+def RLock(*args, **kwargs):
+    """Kazoo read-write lock (reader side).
+
+    Example usage:
+
+    .. code-block:: python
+
+        zk = KazooClient()
```
harlowja: Same issue as above, doesn't exactly look right.

kazoo/recipe/lock.py
```diff
@@ -167,14 +181,15 @@ def _inner_acquire(self, blocking, timeout):
...
+            predecessor = self.predecessor(children, our_index)
+            if not predecessor:
                 return True
 
             if not blocking:
                 return False
 
             # otherwise we are in the mix. watch predecessor and bide our time
-            predecessor = self.path + "/" + children[our_index - 1]
+            predecessor = self.path + "/" + predecessor
```
harlowja: While your at it how about u use the following:

https://github.com/python-zk/kazoo/blob/master/kazoo/protocol/paths.py#L19

harlowja: Good point, let's wait for those.

harlowja:
Have u seen https://github.com/python-zk/kazoo/pull/141 (might be an intersting thing to look at for u).

kazoo/recipe/lock.py
```diff
@@ -43,18 +43,31 @@ class Lock(object):
...
     acquired will raise a ``RuntimeError``.
 
+    This is an exclusive lock. For a read/write lock, see :method:`WLock` and
+    :method:`RLock`.
+
     """
-    _NODE_NAME = '__lock__'
 
-    def __init__(self, client, path, identifier=None):
+    def __init__(self, client, path, identifier=None, node_name="__lock__",
```
harlowja: I'd almost rather have ``class Lock(object):`` -> ``class _Lock(object):``

And then only have these keyword arguments on ``_Lock``; what do u think? Even though u are saying 'don't pass them' it seems like we could do better...

cpatulea:
#141 looks nice, I would be happy to use it.

Are you willing to merge it?

harlowja:
#141 could be combined with yours; anything useful from it that u might want?

cpatulea:
My application code can work as well on top of #141 or this pull request.

I'm also happy to make some minor changes as needed to either #141 or this pull request, so that they can be merged.

However, I feel that this pull request is getting sidetracked. I sent the pull request nearly a week ago, I've addressed all review comments so far, and you are now suggesting some kind of merging of the pull requests, which would involve significant additional work. I am happy to help improve kazoo, but at this point either pull request seems like a good candidate for merging, without additional revision. Any further work can be done later, by whoever depends on that work.

harlowja:
Fair enough, I didn't mean to discredit your code or anything; was just offering some ideas...

OpenSource takes time (and patience); but the more you participate the better it gets :)

kazoo/recipe/lock.py
```diff
@@ -195,8 +218,14 @@ def _get_sorted_children(self):
         children = self.client.get_children(self.path)
 
         # can't just sort directly: the node names are prefixed by uuids
-        lockname = self._NODE_NAME
-        children.sort(key=lambda c: c[c.find(lockname) + len(lockname):])
+        def _seq(c):
```
harlowja: It'd be nice to have a comment here ;)

cpatulea: Done.

harlowja:
Ok let's get this in, further tweaks can be made after merge I think :+1:

kazoo/recipe/lock.py
```diff
@@ -267,6 +303,56 @@ def __exit__(self, exc_type, exc_value, traceback):
         self.release()
 
 
+def WLock(*args, **kwargs):
```
bbangert: Why aren't these subclasses as is typical in Python to slightly modify behavior? The lack of subclassing here means that when you look at tracebacks and such, there's nothing indicating WLock vs RLock vs Lock. That will drive people crazy.

bbangert:
This recipe unfortunately creates Lock objects that are actually Read or Write Locks, without anything in the class object to easily identify them as such. Given the scope of the changes, it definitely needs a thorough set of tests as well which don't seem to be included.

jeblair:
I have continued work on this PR in #419.

bbangert:
Ok, closing this one then.

PR #307, opened 2015-04-09
Re-set the ping_outstanding event before connecting

kormat:
If the client disconnects with the ping_outstanding event set, when it
reconnects the first timeout will cause a "dropped" heartbeat, and make
the library reconnect yet again.


harlowja:
Seems ok to me.

PR #326, opened 2015-05-28
Ensure tox is always installed into the virtualenv.

hannosch:
I tried to run make && make test and it failed for me, as I don't have tox installed in my global environment. This makes sure we install tox into the kazoo virtualenv and use it from there via bin/tox.


hannosch:
This doesn't seem to work, travis-ci builds can't find bin/tox ... revoking PR.

PR #327, opened 2015-05-28
Make use of the new travis-ci caching mechanism.

hannosch:
This moves the Zookeeper install from `bin/zookeeper-3.4.6` to `zookeeper/3.4.6` to make it easier to cache it.

In the new sudo-less environment you can't use `sudo apt-get` anymore. This might result in failing to installed gevent 0.13.8, gevent 1.0.x includes a copy of libev inside the tar.gz, so it should build fine.

First test to see if the caches get populated...

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,6 +24,7 @@ man/
 parts/
 share/
 site-packages/
+zookeeper/
 .coverage
 .idea
 .pip_cache
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,10 @@
+sudo: false
+
+cache:
+  directories:
+    - $HOME/.cache/pip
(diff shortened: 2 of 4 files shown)


hannosch:
Got libevent-dev installed again via the apt addon.

The build and cache are working, there are just individual test failures as before. The cache itself doesn't seem to help in terms of test runtime, though it does make the build less dependent on the Apache mirror for getting the ZK archive.

Since the old sudo infrastructure is going away at some point, it's probably still worth it to switch and merge this.

harlowja:
Looks good to me, although I've been unable to get to figure out the cause of travis still messing up (with nearly always noautherrors); can't seem to cause it to happen locally so unsure whats up with that :-P

Issue #328, opened 2015-05-28
New tox based tests don't install gevent


hannosch:
Since we switched to the tox test runner, we don't run any of the gevent tests anymore, since gevent isn't getting installed into the tox environment.

This used to work via:

`.travis.yml` sets an environment variable like `GEVENT_VERSION=0.13.8` (or 1.0.1).

The Makefile adds another check and figures out the Python version, and sets `GREENLET_SUPPORTED` accordingly. Inside the `make build` section, if greenlet is supported (not PyPy / not Python 3), the gevent dependencies get installed and finally gevent gets installed with a version restriction to GEVENT_VERSION.

But after the switch to tox, `make build` isn't used on travis anymore, but instead tox.ini is responsible for installing dependencies. This does install the gevent / eventlet dependenies, but not gevent of the specified version itself.

The tests `succeed` everywhere, since all the gevent based tests are skipped if it isn't available.

@harlowja any ideas here?

harlowja:
Seems like we just need to have a few variations of gevent in tox.ini env(s) for 1.0.1 and 0.13.8 and then have those tested in travis, will get that fixed and up and running.

PR #330, opened 2015-05-28
Fix up tox envs and their usage in travis

harlowja:
Ensure that we have specific virtualenvs listed in tox that install the
desired gevent and eventlet versions and ensure that the
``.travis.yml`` file references them for usage.

This also adds gevent (with no version restriction) to the py27, py26
builds (with postfix ``-gevent``) and also adds eventlet specific virtualenvs
(with postfix ``-eventlet``) for making sure that eventlet and gevent
continues to work.

Fixes issue #328

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,23 +6,31 @@ matrix:
         - python: "2.7"
     include:
         - python: "2.6"
-          env: GEVENT_VERSION=0.13.8 ZOOKEEPER_VERSION=3.3.6 TOX_VENV=py26
+          env: ZOOKEEPER_VERSION=3.3.6 TOX_VENV=py26
         - python: "2.6"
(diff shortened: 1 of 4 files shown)


hannosch:
Not quite related to this, but I'd probably simplify this by only testing each Python version with gevent==0.13.8 and once with gevent without a version restriction / latest. Since we don't do any of the gevent monkey patching, I wouldn't even run the tests without gevent installed.

Especially since gevent 1.0.2 was just released, and updating the gevent version is a bit tedious.

And to go even further off-topic, we might just stop testing with gevent 0.13.x. The 1.0 line has had a stable release for 1.5 years now and a really good release candidate for 2.5 years.

harlowja:
I'm fine with all the mentioned comments :)

Issue #234, opened 2014-08-11
Add support for reconfig cluster membership operation


hannosch:
ZK 3.5 introduced a new reconfig operation to change the cluster memberships at runtime. The feature was tracked in https://issues.apache.org/jira/browse/ZOOKEEPER-107. The bulk of the work happened in http://svn.apache.org/viewvc?view=revision&revision=1453693.

This is the new type code 16.

rgs1:
WIP: https://github.com/rgs1/kazoo/tree/support-reconfig, comments welcomed! Will add tests later this week.

rgs1:
So this PR has basic coverage for this feature: https://github.com/python-zk/kazoo/pull/333. That being said, there are some server-side bugs to be addressed (i.e.: https://issues.apache.org/jira/browse/ZOOKEEPER-2202), but it would be nice to have this in so I can support it in [zk-shell](https://github.com/rgs1/zk_shell)!

jeffwidman:
It appears this was handled by #333

PR #333, opened 2015-06-01
Add support for reconfig cluster membership operation (issue #234)

rgs1:
Signed-off-by: Raul Gutierrez S <rgs@itevenworks.net>

### DIFF
--- a/kazoo/tests/test_client.py
+++ b/kazoo/tests/test_client.py
@@ -15,6 +15,7 @@ from kazoo.testing import KazooTestCase
 from kazoo.exceptions import (
     AuthFailedError,
     BadArgumentsError,
+    BadVersionError,
     ConfigurationError,
     ConnectionClosedError,
     ConnectionLoss,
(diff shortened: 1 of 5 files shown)


rgs1:
this passes for me locally fwiw (given the tests aren't happy in travis)

kazoo/tests/test_client.py
```diff
@@ -1147,3 +1148,57 @@ def test_unchroot(self):
...
+
+class TestReconfig(KazooTestCase):
+
+    def setUp(self):
+        KazooTestCase.setUp(self)
+        if TRAVIS_ZK_VERSION:
+            version = TRAVIS_ZK_VERSION
+        else:
+            version = self.client.server_version()
+        if not version or version < (3, 5):
```
harlowja: Should this kind of check be denoted in the reconfig docstring? (or should that function even do that check itself?)

rgs1: you mean, checking if the server supports it? you'd just UnimplementedError in that case

harlowja: Maybe that, or at least docstrings?

Yasumoto:
This is really exciting- thanks for adding it!

PR #335, opened 2015-06-09
On demand bind for recipe classes to the client class

nordaux:
Bind all recipes at the client class only once at the class creation time.
Implemented over class decorator, because it's more portable than metaclass.
Also this type of bind saves access to the **doc** attribute of original classes,
what is useful for example with IPython.

### DIFF
--- a/kazoo/utils.py
+++ b/kazoo/utils.py
@@ -1,0 +1,19 @@
+from types import MethodType
+
+
+def _on_demand(cls):
+    def rebind(dct):
+        for name, obj in dct.items():
+            try:
+                wrap, obj = obj
+                if wrap is MethodType:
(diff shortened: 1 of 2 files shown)


kazoo/utils.py
```diff
@@ -0,0 +1,19 @@
...
+
+
+def _on_demand(cls):
+    def rebind(dct):
+        for name, obj in dct.items():
+            try:
+                wrap, obj = obj
+                if wrap is MethodType:
+                    setattr(cls, name, wrap(obj, None, cls))
+            except (TypeError, ValueError):
```
harlowja: Would it be  better to just let this raise (instead of silencing errors?)

bbangert:

PR #336, opened 2015-06-12
Bump version up to 2.2

harlowja:
Version 2.1.1 was never released (due to **repeated** travis build errors)
so we can remove it from being a version number and go directly
to 2.2 (and collapse the CHANGES.rst to account for that).


vilobhmm:
+1

rgs1:
+1 (fwiw)

PR #341, opened 2015-06-17
ignore error when removing directory

tevino:
If I use `destroy` or `reset` after starting the cluster, the error occurs at exit, so I have to restart the closed cluster before exit just to avoid the error.


harlowja:
Can you open an issue to describe what error you see.

tevino:
@harlowja #343

harlowja:
Thanks.

PR #342, opened 2015-06-18
Use 'isinstance' instead of 'type'

harlowja:
It doesn't seem recommended to do `type(X) is Y` when isinstance does this in a more proper manner.


bbangert:

PR #344, opened 2015-06-19
Use testtools for better unittest testcase compatability

harlowja:
Testtools provides compat. methods that make missing unittest
methods on older versions of python exist, so that they can
be used on those older versions.


bbangert:

PR #345, opened 2015-06-19
Ensure we track all threads spawned and join when stopped

harlowja:


bbangert:

PR #346, opened 2015-06-22
Making version number programatically accessible

mattdodge:
This has been documented in a number of PEPs (e.g [PEP 396](https://www.python.org/dev/peps/pep-0396/), [PEP 440](https://www.python.org/dev/peps/pep-0440/)). It essentially makes the following operation possible

```bash
>>> import kazoo
>>> kazoo.__version__
'2.2.1'
```

Only real impact I see is that upon a new version the `kazoo/__init__.py` file would need to be edited, rather than the `setup.py` file like before.


harlowja:
What about a ``kazoo/version.py`` file, is that more natural?

mattdodge:
I've seen that used in practice before, there are lots of different ways to make it work. You would probably have a better feel for the organization of the repository than I would, if that sounds better to you I can easily switch it to that approach. Thoughts?

harlowja:
I'd be fine with a ``version.py`` file, although perhaps thats not the best way (I've used that kind of file before); then u don't need the regex stuff in setup.py and can just use ``from kazoo import version`` ?

mattdodge:
When in doubt, I look at the popular repositories to see how they do it:

1. [Requests](https://github.com/kennethreitz/requests) - In `__init__.py`
2. [Boto](https://github.com/boto/boto) - In `__init__.py` but imported from `setup.py` instead of read from the file directly
3. [Numpy](https://github.com/numpy/numpy) - In `version.py`

So basically, looks like any of them are acceptable

harlowja:
Ya, the repos I have created/maintain, usually a ``version.py``, but not always (sometimes I just have it set in ``setup.py`` as a string...)

mattdodge:
The problem with imports like that is they often don't work when installing on a fresh machine, at least in my experience. That's why I've always used the regex hack

mattdodge:
Ok, I'll switch it over to a `version.py`

harlowja:
Hmmm, I wonder if that's something I've seen, maybe not anymore due to wheels. Whatever floats your boat I guess :-P

harlowja:
Seems ok to me.

PR #350, opened 2015-08-05
Automatically generate the version number for docs

deejay1:
Reuse the code from setup.py to generate the proper version
number for the Sphinx documentation.


docs/conf.py
```diff
@@ -63,10 +64,15 @@ def __getattr__(self, name):
 # |version| and |release|, also used in various other places throughout the
 # built documents.
 #
-# The short X.Y version.
-version = '2.0'
 # The full version, including alpha/beta/rc tags.
-release = '2.0.1'
+release = ''
+here = os.path.abspath(os.path.dirname(__file__))
```
harlowja: I think in this code we can just do the following?

```
  >>> from kazoo import version
  ...
```

The kazoo code is going to be needed to link the docs to the actually classes anyway, so I think we should be fine just importing it here (at least this is how I've commonly seen it done in other projects).

deejay1: Good catch, fixed in next commit.

PR #351, opened 2015-08-05
Fix examples in recipes

deejay1:
Fix the recipe examples, so they actually work by connecting
to ZooKeeper. Without start() they just hang and can't be killed.


harlowja:
:+1:

harlowja:
Thanks!

PR #353, opened 2015-08-18
Make lock._inner_acquire robust to paths without leading slashes.

### LINKED ISSUE #288: retry failure in Lock.acquire() may lead to an AttributeError
I've hit on the following exception, on at least one occasion:

.../lib/python2.7/site-packages/kazoo/recipe/lock.py:123, self._delete_node(self.node)", "'AttributeError: 'Lock' object has no attribute 'node'

Calling Lock.acquire(), whilst unconditionally raising RetryFailedError under Retry.__call__() reproduces the bug.

sasmith:
Fixes #288.

Note that `lock._inner_acquire` already has correct behavior when it retries, so this diff should have two effects when paths without leading slashes are used: it should make locks slightly faster to acquire, and it should make very-small-timeouts less likely to result in crashes.

### DIFF
--- a/kazoo/recipe/lock.py
+++ b/kazoo/recipe/lock.py
@@ -143,7 +143,7 @@ class Lock(object):
             node = self.client.create(self.create_path, self.data,
                                       ephemeral=True, sequence=True)
             # strip off path to node
-            node = node[len(self.path) + 1:]
+            node = node.rpartition("/")[-1]
 
         self.node = node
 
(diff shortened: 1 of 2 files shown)


sasmith:
re-triggering tests, since the failure looked unrelated to my change

harlowja:
Should we just add some small utility function to:

https://github.com/python-zk/kazoo/blob/master/kazoo/protocol/paths.py

Something akin to what ``os.path.split`` does?

bbangert:
Fixed in a later PR.

PR #354, opened 2015-08-19
Fix potential leak in DataWatch and ChildrenWatch

d3matt:
If the func passed to DataWatcher or ChildrenWatch is bound to an object it
will keep a reference to the object around which could prevent it from getting
garbage collected.


harlowja:
:+1:  seems ok to me.

PR #358, opened 2015-09-10
Fix handling of retry failures in Lock.acquire

kormat:
(Fixes https://github.com/python-zk/kazoo/issues/288)

- Make sure that self.node always exists.
- Use _best_effort_cleanup() instead of _delete_node() when we didn't
  get the lock.
- Make RetryFailedError use the existing cleanup code.
- Make _best_effort_cleanup use self.node if it's set, and fall back to
  _find_node otherwise.


harlowja:
:+1:  looks ok to me

kormat:
Can someone merge this, please?

harlowja:
Merged ;)

kormat:
Great, thanks :)

Issue #352, opened 2015-08-10
kazoo.recipe.queue.Queue get() block forever


vipjml:
def _inner_get(self):
        if not self._children:
            self._children = self.client.retry(
                self.client.get_children, self.path)
            self._children = sorted(self._children)
        if not self._children:
            return None
        name = self._children[0]
        try:
            data, stat = self.client.get(self.path + "/" + name)
        except NoNodeError:  # pragma: nocover
            # the first node has vanished in the meantime, try to
            # get another one
            raise ForceRetryError()
        try:
            self.client.delete(self.path + "/" + name)
        except NoNodeError:  # pragma: nocover
            # we were able to get the data but someone else has removed
            # the node in the meantime. consider the item as processed
            # by the other process
            raise ForceRetryError()
        self._children.pop(0)
        return data

_inner_get should clear  self._children when NoNodeError raised  . if _children  is not clear,every time retry,_children   is the same,raise NoNodeError,case Queue.get() block forever.

bbangert:
Closed by #359.

PR #359, opened 2015-09-18
Fixed Queue to clear children before retry

patrickotoole:
- this fixes issue with Queue recipe as reported in #352


harlowja:
:+1:  seems ok to me!

PR #363, opened 2015-11-13
Periodic pings

harlowja:


bbangert:
Per #364 and the bug cited, it's not clear that this is needed.

harlowja:
Agreed, closing for now.

PR #367, opened 2015-11-18
remove listener in watchers.ChildrenWatch

yungchin:
I believe that without this, the watch will remain in place forever,
even if it won't really call the callback again.  The inserted line was
copied straight out of watchers.DataWatch, which follows the same
general design as ChildrenWatch.


bbangert:

PR #370, opened 2015-11-30
Set default retry_max_delay to one minute

FedericoCeratto:
Related to #369


harlowja:
Unsure why this is done, this changes the default values for folks that have been using kazoo without much reason. I mean I get the change, but it seems bad to change this on folks, its already configurable so just provide the value that you desire when creating a client.

PR #371, opened 2015-12-06
Add break_lock method to semaphores

SpamapS:
These may end up being broken through external means, so this method is
used for testing how that scenario is handled. It may also prove useful
for administrative tooling.

### DIFF
--- a/kazoo/recipe/lock.py
+++ b/kazoo/recipe/lock.py
@@ -245,6 +245,17 @@ class Lock(object):
     def acquired_lock(self, children, index):
         return index == 0
 
+    def break_lock(self):
+        children = self._get_sorted_children()
+        # remove the first one
+        if not len(children):
+            return False
+        try:
+            self._delete_node(children[0])
+        except NoNodeError:
+            return False
+        return True
+
     def _watch_predecessor(self, event):
         self.wake_event.set()
 
@@ -554,6 +565,10 @@ class Semaphore(object):
             pass
         self.is_acquired = False
         return True
+
+    def break_lock(self):
+        lock = self.client.Lock(self.path, self.data)
+        return lock.break_lock()
 
     def lease_holders(self):
         """Return an unordered list of the current lease holders.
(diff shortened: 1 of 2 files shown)


kazoo/recipe/lock.py
```diff
@@ -555,6 +566,10 @@ def _inner_release(self):
         self.is_acquired = False
         return True
 
+    def break_lock(self):
```
harlowja: Just wondering about this in the semaphore class, since the lock usage inside of this is somewhat different than the lock usage in the main lock class, maybe needs more advanced breaking logic?

SpamapS: I am flying a bit blind here. This works for the case I have for tooz. Beyond that, I'm entirely open to suggestions.

harlowja: No worries I can see about the semaphore part, it's based on leases so I'm
pretty sure breaking it is slightly more complicated.

On Sunday, December 6, 2015, SpamapS <notifications@github.com> wrote:

> In kazoo/recipe/lock.py
> <https://github.com/python-zk/kazoo/pull/371#discussion_r46773149>:
>
> > @@ -555,6 +566,10 @@ def _inner_release(self):
> >          self.is_acquired = False
> >          return True
> >
> > +    def break_lock(self):
>
> I am flying a bit blind here. This works for the case I have for tooz.
> Beyond that, I'm entirely open to suggestions.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/python-zk/kazoo/pull/371/files#r46773149>.
>


-- 
--
facebook.com/jshharlow <http://www.facebook.com/jshharlow>
flickr.com/jshharlow
YIM: jshharlow

kazoo/tests/test_lock.py
```diff
@@ -624,6 +624,23 @@ def test_acquire_after_cancelled(self):
         self.assertTrue(lock.cancelled)
         self.assertTrue(lock.acquire())
 
+    def test_break_lock(self):
+        lock = self.client.Semaphore(self.lockpath)
```
harlowja: Should this test the self.client.Lock ?

bbangert:

Issue #375, opened 2016-01-07
ensure-zookeeper-env.sh not work any more


vipjml:
zookeeper 3.4.6 is not on server any more. So any new submit can't past test.

https://travis-ci.org/python-zk/kazoo/jobs/97795666

I change it to 3.4.7,but still some test can't be passed:
https://travis-ci.org/python-zk/kazoo/builds/99687754

harlowja:
Ok, this should be all fixed now. Thanks for reporting.

PR #377, opened 2016-01-09
Zookeeper 3.4.6 seems no longer existent on http://apache.osuosl.org

harlowja:
Move to using 3.4.7 which is on http://apache.osuosl.org

Fixes issue #375


d3matt:
where are you seeing a 3.4.7?  current is 3.4.6

Issue #309, opened 2015-04-22
lock.acquire() never success when the path has some other nodes


wangst321:
In our case, we have nodes as below:

    "/mynode01/content"
    "/mynode01/targets"

there are operations: delete and update on the "/mynode01". so use the ``Lock`` to make sure only one  can modify the node.

but ``Lock("/mynode01").acquire`` will never success, because the Lock create a node "uuid__lock__" under the node "/mynode01", and the lock node is always bigger than "content" and "targets" after sorted.

the sort method:

    children.sort(key=lambda c: c[c.find(lockname) + len(lockname):])

the lock node name:

    self.prefix = uuid.uuid4().hex + self._NODE_NAME

below is my suggestion:
   
    self.prefix = self._NODE_NAME + uuid.uuid4().hex

and:

    children.sort()

aschokking:
Created a PR for this as we were running into the same issue: https://github.com/python-zk/kazoo/pull/385

bbangert:
The PR needs to be updated and re-opened. Offhand I think it's better to have a lock node just for the lock without mixing other nodes in it. But it seems safe enough to filter as suggested.

PR #385, opened 2016-03-15
Locks should ignore non-lock node children

aschokking:
Noticed that if a lock node has non-lock children the current logic would consider them in the list of candidates for the lock and if they happen to sort higher than the current correct format nobody will ever be able to acquire the lock.

Proposed fix is to filter out any such non-lock child nodes from consideration when sorting the list.

### DIFF
--- a/kazoo/recipe/lock.py
+++ b/kazoo/recipe/lock.py
@@ -254,6 +254,7 @@ class Lock(object):
         # can't just sort directly: the node names are prefixed by uuids
         lockname = self._NODE_NAME
 
+        # only consider children that were created for this lock
         children = list(filter(lambda c: lockname in c, children))
         children.sort(key=lambda c: c[c.find(lockname) + len(lockname):])
         return children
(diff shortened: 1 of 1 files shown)


aschokking:
Addresses issue noted here: https://github.com/python-zk/kazoo/issues/309

aschokking:
The 1 test failure appears to be unrelated to this change and might just be a flaky timeout?

aschokking:
Looks like the test failures are still just flaky and unrelated to this change :(

aschokking:
I'd love to hear from the maintainers at some point, are you waiting for the flaky tests to pass? If so I'm not sure how best to make that happen without just creating a bunch of no-op commits to keep running them until the monte carlo lines up with all green.

bbangert:
Please re-open to rebase and update as needed. I'd be happy to merge this once it's updated.

aschokking:
I looked at the latest code and I think the current sorting logic in there will take care of this (since it's putting entries which contain the lock name above other nodes)

PR #389, opened 2016-04-02
Adds `PatientMultiChildrenWatch` a generalization of `PatientChildrenWatch`

se7entyse7en:
Generalized the `PatientChildrenWatch` in order to support watching children of multiple paths at the same time.


se7entyse7en:
The tests seem to be broken independently from this PR.

PR #391, opened 2016-04-21
Ensure pending watches are not dropped when connection is lost

michielbaird:
This behavior matches the reference implementation
Introduces a NONE event


michielbaird:
Given this failure in travis is spurious how can I rerun it?

bbangert:
Thanks! I was able to coax it through the tests. :)

PR #395, opened 2016-05-14
System call interrupts

jacksontj:
This is a followup to #250

This PR includes 2 patches both fixing more interrupt issues. Basically this is an upstream issue (in python-- http://bugs.python.org/issue20611).

The first patch changes my previous fix (#250) from raising a timeout to retrying-- the issue we see is that the rest of the kazoo code assumes that a timeout is a timeout (which is reasonable). So instead of reworking all timeouts within kazoo, we can simply retry.

The second patch handles the same during connection setup.

### DIFF
--- a/kazoo/handlers/threading.py
+++ b/kazoo/handlers/threading.py
@@ -147,8 +147,8 @@ class SequentialThreadingHandler(object):
         timeout = args[3] if len(args) == 4 else None
         # either the time to give up, or None
         end = (time.time() + timeout) if timeout else None
-        while timeout is None or time.time() < end:
-            if timeout:
+        while end is None or time.time() < end:
+            if end is not None:
                 args = list(args)  # make a list, since tuples aren't mutable
                 args[3] = end - time.time()  # set the timeout to the remaining time
             try:
@@ -161,6 +161,8 @@ class SequentialThreadingHandler(object):
                 if errnum == errno.EINTR:
                     continue
                 raise
+        # if we hit our timeout, lets return as a timeout
+        return ([], [], [])
 
     def socket(self):
         return utils.create_tcp_socket(socket)
--- a/kazoo/handlers/threading.py
+++ b/kazoo/handlers/threading.py
@@ -143,17 +143,24 @@ class SequentialThreadingHandler(object):
             python2atexit.unregister(self.stop)
 
     def select(self, *args, **kwargs):
-        try:
-            return select.select(*args, **kwargs)
-        except select.error as ex:
-            # if the system call was interrupted, we'll return as a timeout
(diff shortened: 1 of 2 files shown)


jacksontj:
Seems that all the tests failed with some gzip error? Sounds unrelated-- esp. since all tests failed with the same error (and I don't get that error locally)

jacksontj:
cc @bbangert @harlowja 
Since you guys helped review my patch last time-- I figure you'll be interested in this one as well.

bbangert:
This looks fine to me, but I'm also stumped by the Travis errors.

jacksontj:
@bbangert do you know of a way to re-trigger the run? From the errors it looks like travis might have had an issue.

bbangert:
@jacksontj yup, triggered it again just now

jacksontj:
This is really odd, when I run the tests locally they seem to pass just fine. Is there some way to get additional debugging info from travis? Or is there someone who maintains it that we could ping?

jacksontj:
@bbangert any ideas who we can talk to?

bbangert:
@jacksontj not offhand. there might have been a few failures before this one though on master, so I need to retry them as well to see when it broke for good

bbangert:
@jacksontj ok, well, I went back to a prior travis test that only had 2 fail, and reran it.... and then they all failed. I'm guessing something on Travis has changed, such that all our tests are now insta-fail.

bbangert:
I've found the issue. The mirror chosen no longer has 3.4.7, I reverted that so that the tests can run. Will be restarting this shortly.

bbangert:
@jacksontj and of course, that fix won't help you until you rebase against master. Can you rebase against master?

bbangert:
@jacksontj looks like there's 1 error reported.

```
======================================================================

ERROR: test_dirty_sock (kazoo.tests.test_connection.TestConnectionHandler)

----------------------------------------------------------------------

  File "/home/travis/build/python-zk/kazoo/kazoo/tests/test_connection.py", line 229, in test_dirty_sock

    wait(lambda: client.handler.select([read_sock], [], [], 0)[0] == [])

  File "/home/travis/build/python-zk/kazoo/kazoo/tests/util.py", line 105, in __call__

    if func():

  File "/home/travis/build/python-zk/kazoo/kazoo/tests/test_connection.py", line 229, in <lambda>

    wait(lambda: client.handler.select([read_sock], [], [], 0)[0] == [])

TypeError: 'NoneType' object is unsubscriptable
```

jacksontj:
@bbangert fixed the issues :) now travis is happy.

bbangert:
@jacksontj looks good!

PR #397, opened 2016-06-10
Counter: expose the previous and changed values.

neuroid:
This commit extends the `Counter` API with the `pre_value` and
`post_value` attributes. These attributes are set when a counter
is changed. Previously, if the counter was modified concurrently,
it was not possible to access the exact value to which the counter
was set after a successful increment or decrement.

### DIFF
--- a/kazoo/recipe/counter.py
+++ b/kazoo/recipe/counter.py
@@ -34,10 +34,14 @@ class Counter(object):
         counter += 2
         counter -= 1
         counter.value == 1
+        counter.pre_value == 2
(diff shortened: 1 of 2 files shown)


neuroid:
This is inspired by the API provided by http://curator.apache.org/curator-recipes/distributed-atomic-long.html

neuroid:
Ping @hannosch

hannosch:
@neuroid sorry, I'm not actively contributing to kazoo anymore.

PR #398, opened 2016-06-22
New recipe proposal: TreeCache

tonyseek:
This is a port of the Apache Curator's TreeCache recipe. It builds an in-memory cache of a subtree in ZooKeeper and keeps it up-to-date.

Please review it. Thanks!

### DIFF
--- a/kazoo/recipe/cache.py
+++ b/kazoo/recipe/cache.py
@@ -248,15 +248,14 @@ class TreeNode(object):
         self._refresh_children()
 
     def _refresh_data(self):
-        self._tree._outstanding_ops += 1
         self._call_client('get', self._path)
 
     def _refresh_children(self):
         # TODO max-depth checking support
-        self._tree._outstanding_ops += 1
         self._call_client('get_children', self._path)
 
     def _call_client(self, method_name, path, *args):
+        self._tree._outstanding_ops += 1
         callback = functools.partial(
             self._tree._in_background, self._process_result,
             method_name, path)
@@ -281,7 +280,8 @@ class TreeNode(object):
         logger.debug('process_result: %s %s', method_name, path)
         if method_name == 'exists':
             assert self._parent is None, 'unexpected EXISTS on non-root'
-            if result.successful():
+            # the value of result will be set with `None` if node not exists.
+            if result.get() is not None:
                 if self._state == self.STATE_DEAD:
                     self._state = self.STATE_PENDING
                 self.on_created()
--- a/kazoo/recipe/cache.py
+++ b/kazoo/recipe/cache.py
@@ -58,7 +58,7 @@ class TreeCache(object):
 
         .. note::
 
-            This method is not thread safety.
+            This method is not thread safe.
         """
         if self._state == self.STATE_LATENT:
             self._state = self.STATE_STARTED
@@ -79,7 +79,7 @@ class TreeCache(object):
 
         .. note::
 
-            This method is not thread safety.
+            This method is not thread safe.
         """
         if self._state == self.STATE_STARTED:
             self._state = self.STATE_CLOSED
@@ -155,7 +155,7 @@ class TreeCache(object):
     def _publish_event(self, event_type, event_data=None):
         event = TreeEvent.make(event_type, event_data)
         if self._state != self.STATE_CLOSED:
-            logger.debug('public event: {0!r}'.format(event))
+            logger.debug('public event: %r', event)
             self._in_background(self._do_publish_event, event)
 
     def _do_publish_event(self, event):
@@ -260,7 +260,7 @@ class TreeNode(object):
         method(path, *args, **kwargs).rawlink(callback)
 
     def _process_watch(self, watched_event):
-        logger.debug('process_watch: {0!r}'.format(watched_event))
+        logger.debug('process_watch: %r', watched_event)
         with handle_exception(self._tree._error_listeners):
             if watched_event.type == EventType.CREATED:
                 assert self._parent is None, 'unexpected CREATED on non-root'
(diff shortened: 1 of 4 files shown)


harlowja:
Neat 👍

tonyseek:
@bbangert Would you help us to review this new recipe proposal?

The TreeCache recipe is necessary for building high-availability service discovery middlewares. It provides eventually consistent snapshots of a ZooKeeper ensemble. Unlike `DataWatch` and `ChildrenWatch`, the TreeCache handles entire subscribed subtree instead of a single layer.

Our team (@eleme) have deployed this recipe to a part of machines in our production environment. I (@tonyseek), Haochuan Guo (@wooparadog) and Tianwen Zhang (@tevino) are maintaining the code.

bbangert:
Sure, it'll be a week before I can get to it though as I'm on a company
retreat till then.

kazoo/recipe/cache.py
```diff
@@ -0,0 +1,382 @@
...
+        The cache is not started automatically. You must call this method.
+
+        After a cache started, all changes of subtree will be synchronized
+        from the ZooKeeper server. Events will be fired for those activity.
+
+        See also :meth:`~TreeCache.listen`.
+
+        .. note::
+
+            This method is not thread safety.
```
harlowja: NIt: thread safe

kazoo/recipe/cache.py
```diff
@@ -0,0 +1,382 @@
...
+        """
+        if self._state == self.STATE_LATENT:
+            self._state = self.STATE_STARTED
+        else:
+            raise RuntimeError('already started')
```
harlowja: Raise a `KazooException` instead of a `RuntimeError`?

kazoo/recipe/cache.py
```diff
@@ -0,0 +1,382 @@
...
+        return cls((path, data, stat))
+
+
+@contextlib.contextmanager
+def handle_exception(listeners):
+    try:
+        yield
+    except Exception as e:
+        logger.debug('processing error: {0!r}'.format(e))
+        for listener in listeners:
```
harlowja: This seems to drop the exception if there are no listeners, that seems bad, can u add in the logic to not drop it.

kazoo/recipe/cache.py
```diff
@@ -0,0 +1,382 @@
...
+        :returns: A :class:`~kazoo.recipe.cache.NodeData` instance.
+        """
+        return cls((path, data, stat))
+
+
+@contextlib.contextmanager
+def handle_exception(listeners):
+    try:
+        yield
+    except Exception as e:
```
harlowja: I'd also rather not catch `Exception` here and just catch `KazooException` to start (`Exception` is to broad).

kazoo/recipe/cache.py
```diff
@@ -0,0 +1,382 @@
...
+
+    def was_reconnected(self):
+        self._refresh()
+        for child in self._children.values():
+            child.was_reconnected()
+
+    def was_created(self):
+        self._refresh()
+
+    def was_deleted(self):
```
harlowja: What is this method supposed to do? IMHO a method that starts with 'was' should return a boolean, but these don't seem to do that. A little confused by the naming here.

tevino: @harlowja I believe it was ported from [the original code base](https://git-wip-us.apache.org/repos/asf?p=curator.git;a=blob;f=curator-recipes/src/main/java/org/apache/curator/framework/recipes/cache/TreeCache.java;h=ed32223819fb5d78004142c26f621e7eb6b3052b;hb=refs/heads/CURATOR-3.0#l287).

kazoo/recipe/cache.py
tonyseek: I'm afraid of we could not control which exception user listeners will throw. If we don't catch `Exception`, one listener may stop other listeners to be triggered by throwing an unexpected exception.

kazoo/recipe/cache.py
tonyseek: You are right. I should write logs here if no listeners.

kazoo/recipe/cache.py
harlowja: I think its just the base of all kazoo exceptions, so I think using it should be fine (and adding a docstring saying what will happen).

kazoo/recipe/cache.py
harlowja: Weird, can we change it :)

kazoo/recipe/cache.py
harlowja: Right, sounds like we just need to make the expectations more clear as to what users listeners should and shouldn't do.

tonyseek: If we ask people don't throw any exception except `KazooException` in their listeners, people must remember to write `try: ... except Exception: ...` in all listeners, or they will have unexpected behavior. It looks weird.

I don't like the "catch and hide" style too, especially for unknown exceptions. But we have `logger.exception` here. I believe it will leave enough information for unexpected exceptions from listeners.

tonyseek:
@bbangert @harlowja Revised

tonyseek:
@bbangert @harlowja Request to review 😸

bbangert:
Um, wow! This looks great, thorough docs, API comments, doc strings, and a bunch of tests. Thanks for the great work!

bbangert:
This unfortunately no longer works against master, it causes a RuntimeError in the tests, so I had to revert it. If you could rebase and open a new PR that doesn't cause test halts I'd love to include this.

tonyseek:
👌  I will open a new PR.

Issue #229, opened 2014-08-07
Deadlock in Kazoo 2.0 w/ add_auth() and retry().


diranged:
We discovered a problem with our nd_service_registry code and Kazoo 2.0 that I could use help with.

The problem manifests as a dead-lock when we instantiate a nd_service_registry.KazooServiceRegistry() object with a username/password setting, and then call the set_node() method.

The issue seems to be line 855 (https://github.com/Nextdoor/ndserviceregistry/blob/master/nd_service_registry/__init__.py#L827-L856) where we use the KazooClient.handler.lock_object() method to get a run lock, then call the KazooClient.retry() method on the KazooClient.add_auth() method. If we remove either the 'with self._run_lock' line (827), OR we remove the self._zk_retry() line (855), the problem goes away and our code works just fine.

Any thoughts on whats changed in Kazoo and what we're doing wrong?

Code Snippet: https://github.com/Nextdoor/ndserviceregistry/blob/master/nd_service_registry/__init__.py#L827-L856

Issue #254, opened 2014-10-23
Unhandled Exception in Connection Loop: RuntimeError: ('xids do not match, expected %r received %r', 28, 27)


diranged:
Two days ago we upgraded our servers from Kazoo 1.3.1 -> Kazoo 2.0. We started seeing a 10x increase in the number of connection failures ... and the recovery of those connections is far worse too. It looks like we're seeing a new exception being raised that we did not see before.

```
        "stack_trace": [
            "Traceback (most recent call last):", 
            "  File '/mnt/.venv/lib/python2.7/site-packages/kazoo/protocol/connection.py', line 531, in _connect_attempt", 
            "    response = self._read_socket(read_timeout)", 
            "  File '/mnt/.venv/lib/python2.7/site-packages/kazoo/protocol/connection.py', line 407, in _read_socket", 
            "    return self._read_response(header, buffer, offset)", 
            "  File '/mnt/.venv/lib/python2.7/site-packages/kazoo/protocol/connection.py', line 338, in _read_response", 
            "    'received %r', xid, header.xid)", 
            "RuntimeError: ('xids do not match, expected %r received %r', 28, 27)"
```

This is being caught by `kazoo.client` and throwing the `Unhandled exception in connection loop` error. I'm completely stumped, I can't seem to replicate this in my own dev environment... and it only happens sporadically in production.

jeffwidman:
bump @diranged any updates per above?

Is it still reproducible?

PR #400, opened 2016-07-29
Fix the client.add_auth hangs by xids mismatch.

tonyseek:
This commit fixes https://github.com/python-zk/kazoo/issues/229  by throwing the runtime exception into the dequeued `async_result`. It will end the waiting of user threading.

But this commit doesn't fix the xids mismatch itself (https://github.com/python-zk/kazoo/issues/254). The unordered xids may caused by a bug from the ZooKeeper server side, such as the official issues [ZOOKEEPER-1863](https://issues.apache.org/jira/browse/ZOOKEEPER-1863) described.

### DIFF
--- a/kazoo/protocol/connection.py
+++ b/kazoo/protocol/connection.py
@@ -338,8 +338,10 @@ class ConnectionHandler(object):
         if header.zxid and header.zxid > 0:
             client.last_zxid = header.zxid
         if header.xid != xid:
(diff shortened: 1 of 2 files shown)


tonyseek:
@bbangert Could you review this patch?

CC @wooparadog

bbangert:
Looks good, nice test too!

Issue #383, opened 2016-02-15
KazooTimeoutError does not inherit from KazooException


jgehrcke:
Is there a specific reason for `KazooTimeoutError` inheriting directly from `Exception` instead of from `KazooException`?

PR #402, opened 2016-08-11
#383 KazooTimeoutError should inherit kazooException

Idem:
leaving `kazoo.protocol.connection.RWServerAvailable` where it is since it's more a logical event handler than a KazooException.

### DIFF
--- a/docs/api/handlers/eventlet.rst
+++ b/docs/api/handlers/eventlet.rst
@@ -16,3 +16,5 @@ Private API
 
   .. autoclass:: AsyncResult
      :members:
+
+  .. autoexception:: KazooTimeoutError
--- a/docs/api/handlers/threading.rst
+++ b/docs/api/handlers/threading.rst
@@ -17,4 +17,4 @@ Private API
   .. autoclass:: AsyncResult
      :members:
 
-  .. autoexception:: TimeoutError
+  .. autoexception:: KazooTimeoutError
--- a/kazoo/exceptions.py
+++ b/kazoo/exceptions.py
@@ -43,7 +43,11 @@ class WriterNotClosedException(KazooException):
     """
 
 
-class KazooTimeoutError(Exception):
+class TimeoutError(Exception):
+    pass
+
+
+class KazooTimeoutError(Exception, TimeoutError):
     pass
 
 
--- a/kazoo/exceptions.py
+++ b/kazoo/exceptions.py
@@ -47,6 +47,21 @@ class KazooTimeoutError(Exception):
     pass
 
 
+class ForceRetryError(Exception):
+    """Raised when some recipe logic wants to force a retry."""
+
+
+class RetryFailedError(KazooException):
+    """Raised when retrying an operation ultimately failed, after
+    retrying the maximum number of attempts.
+    """
+
+
+class InterruptedError(RetryFailedError):
+    """Raised when the retry is forcibly interrupted by the interrupt
+    function"""
+
+
 def _invalid_error_code():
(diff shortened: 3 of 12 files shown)


Idem:
I got the travis tests working on my fork https://travis-ci.org/Idem/kazoo/builds/151559078 but I had to restart some of the jobs (some timeout happened from time to time).

Can anyone with the proper wrights restart the failing jobs?

Idem:
@bbangert Can you review this and trigger new travis build plz?

kazoo/client.py
```diff
@@ -9,46 +9,39 @@
...
-)
+from kazoo.exceptions import AuthFailedError
+from kazoo.exceptions import ConfigurationError
+from kazoo.exceptions import ConnectionClosedError
+from kazoo.exceptions import ConnectionLoss
+from kazoo.exceptions import KazooException
+from kazoo.exceptions import NodeExistsError
+from kazoo.exceptions import NoNodeError
+from kazoo.exceptions import SessionExpiredError
+from kazoo.exceptions import WriterNotClosedException
```
bbangert: This is not the desired style for consistency in the kazoo code-base.

kazoo/client.py
```diff
@@ -9,46 +9,39 @@
...
+from kazoo.protocol.serialization import Exists
+from kazoo.protocol.serialization import GetACL
+from kazoo.protocol.serialization import GetChildren
+from kazoo.protocol.serialization import GetChildren2
+from kazoo.protocol.serialization import GetData
+from kazoo.protocol.serialization import Reconfig
+from kazoo.protocol.serialization import SetACL
+from kazoo.protocol.serialization import SetData
+from kazoo.protocol.serialization import Sync
+from kazoo.protocol.serialization import Transaction
```
bbangert: Not sure why these are all being written in this style now.

bbangert:
I don't understand why all the imports are being rewritten into this new style.

I do agree with the fundamental of this PR though, which is that the TimeoutError should inherit from KazooException so that someone catching Kazoo related exceptions can catch them all properly.

This changes seems problematic since it could break a lot of external code assuming it will get a TimeoutError. Perhaps the KazooTimeoutError should inherit from the original TimeoutError as well so that it's still caught by existing code.

kazoo/client.py
Idem: Will remove this side changes.

Idem:
Thanks for your feedback @bbangert

- [ ] I'll put the previous import style back (Sorry for that. Habits from other projects).
- [ ] `KazooTimeoutError` will now inherit from `TimeoutError` to preserve compatibility. That makes sense.
- [ ] I'll move `TimeoutError` to `kazoo.exceptions` with others exceptions and still import in `kazoo.handler.eventlet` (avoiding cross import, and ensuring compatibility for scripts importing `kazoo.handler.eventlet.TimeoutError`).

I can also drop the 2 last points and leave the `TimeoutError` as it is in `kazoo.handler.eventlet`, if you prefer since it's outside of the scope of the issue.

Idem:
Tests are passing on my forked branch (https://travis-ci.org/Idem/kazoo/builds/157704525)

jgehrcke:
Explicitly linking this to #383 :-).

harlowja:
Weird, why isn't this tested by travis?

bbangert:
Looks good, though for some reason the tests aren't running on it. Can you squash the commits and re-open the PR?

Idem:
Done.
See https://github.com/python-zk/kazoo/pull/447

PR #404, opened 2016-08-23
fix xid out of range

luofeilong:
fix xid out of range (-2147483648 <= number <= 2147483647)

### DIFF
--- a/kazoo/protocol/connection.py
+++ b/kazoo/protocol/connection.py
@@ -439,7 +439,7 @@ class ConnectionHandler(object):
         if request.type == Auth.type:
             xid = AUTH_XID
         else:
-            self._xid += 1
+            self._xid = (self._xid % 2147483647) + 1
             xid = self._xid
 
         self._submit(request, connect_timeout, xid)


kazoo/protocol/connection.py
```diff
@@ -439,7 +439,7 @@ def _send_request(self, read_timeout, connect_timeout):
         if request.type == Auth.type:
             xid = AUTH_XID
         else:
-            self._xid += 1
+            self._xid = (self._xid % 2147483647) + 1
```
harlowja: Can u add a comment as to why u picked this number; and a link ideally to the zookeeper docs so that if it ever changes we can know. Doesn't it also rollover to -2147483648 instead of zero (I though this was a signed int in java)

luofeilong: kazoo/protocol/connection.py, line 276
b.extend(int_struct.pack(xid))

xid += 1 in python can be bigger than 2147483648, but int_struct.pack(xid) just pack xid to int of 4 bytes, so xid must  in range -2147483648~2147483647

harlowja: Right, but doesn't this loose the negative portion?

luofeilong: Only a few digits in the negative portion are used, such as -1, -2
Normal communication using only positive

harlowja: Ok.

PR #407, opened 2016-10-18
Manage session expirations with connection timeout

ralt:
Before this change, kazoo is getting a TimeoutError (transformed into a
ConnectionDropped error by the _socket_error_handling manager), which
doesn't let the zk loop know that it should stop connecting, even long
after the session has expired. It means that kazoo was forever looping,
trying to get a new connection.

### DIFF
--- a/kazoo/client.py
+++ b/kazoo/client.py
@@ -195,6 +195,7 @@ class KazooClient(object):
         self.read_only = read_only
 
         if client_id:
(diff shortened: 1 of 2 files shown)


ralt:
Sorry, this is against the ZK protocol as defined here: https://zookeeper.apache.org/doc/trunk/zookeeperProgrammers.html#ch_zkSessions

Closing.

PR #409, opened 2016-10-20
Let the client keep the session open when closing

ralt:
The use case is being able to restart a process (e.g. software upgrade)
and not losing the ephemeral nodes. The process can save the session id
and password in a file, reload them at boot time, and load the kazoo
client with the previous session id.

### DIFF
--- a/kazoo/client.py
+++ b/kazoo/client.py
@@ -108,7 +108,7 @@ class KazooClient(object):
                  timeout=10.0, client_id=None, handler=None,
                  default_acl=None, auth_data=None, read_only=None,
                  randomize_hosts=True, connection_retry=None,
-                 command_retry=None, logger=None, **kwargs):
+                 command_retry=None, logger=None, keep_session=False, **kwargs):
         """Create a :class:`KazooClient` instance. All time arguments
         are in seconds.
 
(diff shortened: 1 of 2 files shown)


kazoo/client.py
```diff
@@ -594,7 +597,10 @@ def stop(self):
             return
 
         self._stopped.set()
-        self._queue.append((CloseInstance, None))
+        if self._keep_session:
+            self._queue.append((None, None))
```
harlowja: Would it be better to use another named constant, instead of None here, something a little more obvious as to what this is doing?

bbangert:
This will need some tests before it can be merged, ideally something in the docs about it as well. I like the concept over all and think it's good for a new PR with tests/docs.

jeffwidman:
@bbangert If you think this is a good idea, why did you close the PR? 

Normal etiquette on all the OSS projects I'm involved with is to at least ping the original PR author and ask if they're interested in updating it and only after a period of further radio silence to close it for inactivity.

bbangert:
Because GitHub provides no usable UX for me to track effectively which PR's had pings, which had no response, and which were updated. When over 50+ PR's are in the backlog, the entire repo unfortunately becomes unusable to an extent. I apologize for the close, as I was using it as a bad UX hack on GH so that I could see updated/new PR's ready for review.

ralt:
The close was fine by me. It's not unusual to close a PR that hasn't gotten an update in 6 months, specifically stating that the feature would be accepted if the author is still interested in the involved work.

PR #419, opened 2017-03-10
Implement read-write (shared) lock from ZooKeeper recipe

jeblair:
This is a continuation of https://github.com/python-zk/kazoo/pull/306

It addresses the remaining review comments and adds documentation.

### DIFF
--- a/kazoo/recipe/lock.py
+++ b/kazoo/recipe/lock.py
@@ -274,7 +274,14 @@ class Lock(object):
     def _get_sorted_children(self):
         children = self.client.get_children(self.path)
 
-        # can't just sort directly: the node names are prefixed by uuids
+        # Node names are prefixed by uuids: strip the prefix first, which may
+        # be one of multiple values in case of a read-write lock, and return
+        # only the sequence number (as a string since it is padded and will sort
+        # correctly anyway).
+        #
+        # In some cases, the lock path may contain nodes with other prefixes
+        # (eg. in case of a lease), just sort them last ('~' sorts after all
+        # ASCII digits).
         def _seq(c):
             for name in ["__lock__", "__rlock__"]:
                 idx = c.find(name)
--- a/kazoo/recipe/lock.py
+++ b/kazoo/recipe/lock.py
@@ -274,7 +274,7 @@ class Lock(object):
     def _get_sorted_children(self):
         children = self.client.get_children(self.path)
 
-        # Node names are prefixed by uuids: strip the prefix first, which may
+        # Node names are prefixed by a type: strip the prefix first, which may
         # be one of multiple values in case of a read-write lock, and return
         # only the sequence number (as a string since it is padded and will sort
         # correctly anyway).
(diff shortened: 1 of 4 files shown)


jeblair:
The only travis failures are because it's failing to install gevent for py26, which is unrelated to this change.

kazoo/recipe/lock.py
```diff
@@ -323,6 +350,64 @@ def __exit__(self, exc_type, exc_value, traceback):
...
+class ReadLock(Lock):
+    """Kazoo Read Lock
+
+    Example usage with a :class:`~kazoo.client.KazooClient` instance:
+
+    .. code-block:: python
+
+        zk = KazooClient()
+        zk.start()
+        lock = zk.WriteLock("/lockpath", "my-identifier")
```
Shrews: s/WriteLock/ReadLock/

kazoo/recipe/lock.py
```diff
@@ -242,18 +253,34 @@ def _inner_acquire(self, blocking, timeout):
             finally:
                 self.client.remove_listener(self._watch_session)
 
-    def acquired_lock(self, children, index):
-        return index == 0
+    def predecessor(self, children, index):
+        for c in children[:index]:
+            if any(n in c for n in self._EXCLUDE_NAMES):
```
harlowja: Just wanted to check, is this logic right, it seems to be only finding children that have names in exclude names, should it be not in ?

harlowja:
@bbangert can u look this over as well?

kazoo/recipe/lock.py
jeblair: The predecessor function returns the first node ahead of us in
the list which is mutually exclusive with the kind of lock we are
trying to acquire.  So read it as "names which are exclusive with
us" rather than "names to exclude from the list".

In the case of a write lock, the exclude list includes "rlock"
and "lock", so this will return the first of either of those
types -- meaning that if there is a client ahead of us with
either a read or write lock, we have not acquired the lock and
must wait for that client to release it (they are a predecessor).

In the case of a read lock, the exclude list includes
only "lock", so this will return the first write lock entry
only (our predecessor in that case).  It will ignore any read
lock entries ahead of us.

So I believe the logic is correct.  I will note, however, that
the current code will set the watch on the last node which
preceded our node.  This is efficient because if there are N
predecessors, then we would only set the watch on the latest and
would therefore only wake up and go through the loop after all
the predecessors had released their locks.  This patch, however,
sets the watch on the first node which precedes ours, meaning
that we will wake up and iterate through the loop for all N
predecessor nodes.  I'll push up a patch to correct that.

kazoo/recipe/lock.py
jeblair: Fixed.

pabelanger:
Hello! If would be super awesome if we could land this. Anything I can help with?

bbangert:
LGTM! Merging.

Issue #226, opened 2014-08-05
Single watch for multiple nodes can't detect node path


AlexanderplUs:
When I create a single watch for multiple nodes

``` python
input_items = zk.get_children( inputNode )
for item_id in input_items :
    @zk.ChildrenWatch( '{}/{}'.format( inputNode, item_id ) )
    def watch_item_children( items):
        print items
```

my watch callback can't detect for which node it has been called. As I can see from sources of ChildrenWatch class there is _path_ property but it isn't sent to watch callback as a parameter.
I try to use _send_event=True_ and get path from _event_ param but it is set to _None_ at first call
So it will be great if you add additional flag for ChildrenWatch (_send_path_ for example) to send path for watch callback

PR #421, opened 2017-03-16
Process multiple nodes with one watch function

AlexanderplUs:
Added send_path parameter for ChildrenWatch and DataWatch  which allows to process multiple nodes with one watch function as it will recieve node path parameter on call. See  #226


bbangert:
Looks good, but will need some tests first. I'd be happy to merge when those are added, you can open a new PR or re-open this one after rebasing. Thanks!

PR #422, opened 2017-03-16
Client is now context aware

brianbruggeman:
Ran into an issue today where ZK was struggling and this was partially due to the unexpected behavior surrounding .stop vs .close.

This change allows the client to be used as a context and does the "right" thing for cleanup.

Thanks for considering!

### DIFF
--- a/kazoo/client.py
+++ b/kazoo/client.py
@@ -294,6 +294,13 @@ class KazooClient(object):
             raise TypeError('__init__() got unexpected keyword arguments: %s'
                             % (kwargs.keys(),))
 
+    def __enter__(self):
+        return self
+
+    def __exit__(self):
+        self.stop()
+        self.close()
+
     def _reset(self):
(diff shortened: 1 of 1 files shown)


kazoo/client.py
```diff
@@ -294,6 +294,13 @@ def _retry(*args, **kwargs):
             raise TypeError('__init__() got unexpected keyword arguments: %s'
                             % (kwargs.keys(),))
 
+    def __enter__(self):
+        return self
+
+    def __exit__(self):
```
harlowja: Fix this guys signature.

https://docs.quantifiedcode.com/python-anti-patterns/correctness/exit_must_accept_three_arguments.html#anti-pattern

brianbruggeman:
Thanks for the catch!

bbangert:
Cool, I like the concept! It does need a test and some documentation in the client docs though. If you can add those and re-open this PR I'd be happy to merge it.

Issue #429, opened 2017-04-20
Semaphore deadlock on nonblocking acquiring


anskornyakov:
I have a bunch of services running thousands instances which using synchronization via kazoo zookeeper semaphore implementation. Network issues on particular instance or net segment is an everyday occurrence and I m facing deadlocks on my semaphores hardly when such issues appears.

The reason of a deadlock is an zombie lock nodes, it typically looks like:
```
(CONNECTED) /some_node/shard-semaphore-__lock__> ls
ecf828622e7d432bb5801267456599fc__lock__0000000001    // zombie lock node
ccf828622e7d432bb5801267456599fc__lock__0000000191
acf828622e7d432bb5801267456599fc__lock__0000000192
dcf828622e7d432bb5801267a56599fc__lock__0000000194
```

Usually zombie node owner is an instance that failed to delete that node due to network issues the go through and trying to acquire semaphore again and creating new lock node, but as soon as zombie node with  less sequence number exists, it or another instance will never be succeed in obtaining semaphore, and that's why here we has a deadlock.

How does it happens? I'm using nonblocking acquiring.  So` blocking=False `here

 ```
#kazoo.recipe.lock.Lock
def acquire(self, blocking=True, timeout=None):
        def _acquire_lock():
            got_it = self._lock.acquire(False)
            if not got_it:
                raise ForceRetryError()
            return True

        retry = self._retry.copy()
        retry.deadline = timeout

        # Ensure we are locked so that we avoid multiple threads in
        # this acquistion routine at the same time...
        locked = self._lock.acquire(False)
        if not locked and not blocking:
            return False
        if not locked:
            # Lock acquire doesn't take a timeout, so simulate it...
            try:
                locked = retry(_acquire_lock)
            except RetryFailedError:
                return False
        already_acquired = self.is_acquired
        try:
            gotten = False
            try:
                gotten = retry(self._inner_acquire,
                               blocking=blocking, timeout=timeout)  # <<<< 1. Assuming we are not blocking and failed to obtain lock, we get False here
            except RetryFailedError:
                pass
            except KazooException:
                # if we did ultimately fail, attempt to clean up
                exc_info = sys.exc_info()
                if not already_acquired:
                    self._best_effort_cleanup()
                    self.cancelled = False
                six.reraise(exc_info[0], exc_info[1], exc_info[2])
            if gotten:
                self.is_acquired = gotten
            if not gotten and not already_acquired:  # <<<< 2. Trying to do best effort cleanup
[...]

PR #430, opened 2017-04-20
fix deadlock in nonblocking semaphore acquiring

anskornyakov:
Issue #429 fix

fix deadlock in nonblocking semaphore acquiring by doing semaphore inner lock release for every lock acquiring attempt


bbangert:
Thanks for the fix!

PR #435, opened 2017-06-01
Let the client keep the session open when closing

ralt:
The use case is being able to restart a process (e.g. software upgrade)
and not losing the ephemeral nodes. The process can save the session id
and password in a file, reload them at boot time, and load the kazoo
client with the previous session id.


ralt:
Following up on #409.

I'm not sure of what needs to be done with regards to the documentation, the existing one for the client looks vastly outdated, and it doesn't feel appropriate to document all of the arguments at once. :)

ralt:
cc @harlowja @bbangert

ralt:
The CI failures look unrelated to this PR:

```
test_keep_session (kazoo.tests.test_client.TestClient) ... ok
```

kazoo/protocol/connection.py
```diff
@@ -567,7 +575,10 @@ def _connect_attempt(self, host, port, retry):
                         response = self._read_socket(read_timeout)
                         close_connection = response == CLOSE_RESPONSE
                     else:
-                        self._send_request(read_timeout, connect_timeout)
+                        close_connection = self._send_request(
+                            read_timeout,
+                            connect_timeout,
+                        ) == SESSION_KEPT
```
bbangert: This feels a bit awkward, tacking on a boolean eval at the end. Can it be changed to get the response first as was done immediately above it?

bbangert:
I understand the use-case, I think the code/test looks good. I'm a bit worried about the implications, since there is no way to restore Lock's/etc or any of the recipes based on existing ephemeral nodes. It means that when the program starts up, if it had Lock nodes out there, they'll remain there, even though there's no way for this client to recover the use of them.

Does your code using this also save information upon quitting about all the ephemeral nodes and where they were? Best practice with Zookeeper has generally been to avoid attempting to recover a session, and if your process needs to exit, it's better to start over.

bbangert:
Regarding the tests, it looks like this change introduces some issues with the way the connection is handled. While the test you added passes, a lot of others no longer pass, but they passed before this change so there's definitely more work to be done as it appears to have unintended side-effects.

ralt:
> Does your code using this also save information upon quitting about all the ephemeral nodes and where they were? Best practice with Zookeeper has generally been to avoid attempting to recover a session, and if your process needs to exit, it's better to start over.

Our code expects to find ephemeral nodes at specific places. If it's not there, we can know that one client or another lost the session, and thus act on this.

> Regarding the tests, it looks like this change introduces some issues with the way the connection is handled. While the test you added passes, a lot of others no longer pass, but they passed before this change so there's definitely more work to be done as it appears to have unintended side-effects.

Hm, ok, I'll look at them. Thanks!

bbangert:
@ralt Did you want to look into fixing the tests so they pass for the others, or did the change introduce something to the connection handling too subtle to pick out easily?

ralt:
@bbangert I want to look into it, just haven't had the time yet, it's a scarce resource :-)

bbangert:
@ralt I should note that it appears #305 introduced a bug with runtime error mismatches. I'm seeing if that can be cleared up, or if I'll have to revert that PR. At which point maybe this one will actually work fine with no changes.

bbangert:
Updated the base branch to trigger a test run, this should help ensure whether #305 was the issue.

bbangert:
Oh, guess not, tests still all fail.

bbangert:
Just a heads-up, we're coming up on 30 days now. I'm trying to avoid having PR's idle eternally in the backlog like this, so if you'd like to try later, a new PR can be opened when ready.

ralt:
Yes, let me close that and I'll open a new PR when I'll have some time to look at the tests.

PR #447, opened 2017-06-02
[fixes #383] ensure every Exception inherits from KazooException

### LINKED ISSUE #383: KazooTimeoutError does not inherit from KazooException

Idem:
Fixing #383


kazoo/handlers/eventlet.py
```diff
@@ -13,6 +13,7 @@
 from eventlet.green import threading as green_threading
 from eventlet import queue as green_queue
 
+from kazoo.exceptions import KazooTimeoutError, TimeoutError
```
Idem: As per https://github.com/python-zk/kazoo/pull/402#issuecomment-244782581 the goal is too allow one to keep importing TimeoutError from this module for compatibility.
If that is ok for you @bbangert , I can add an ignore mark on that error.

bbangert: I think that's fine, as it keep's the API consistent then. You can add ``# noqa`` at the end of the line and it'll pass the flake8.

bbangert:
@Idem Did you want to update this so I can merge it?

Idem:
I was looking into the contributing guidelines and where into the process of signing the commits and co.

I obviously need to add the #noqa for hound and pass the travis-ci.
So yes, I intend to clean this up a little more.

bbangert:
I triggered the build by updating base just to ensure all the tests look good still.

kazoo/handlers/eventlet.py
tonyseek: Another way to suppress the lint is adding a `__all__ = [..., 'TimeoutError']` line.

bbangert:
Just a heads-up, we're coming up on 30 days now. I'm trying to avoid having PR's idle eternally in the backlog like this, so if you'd like to try later, a new PR can be opened when ready. This is a fairly basic change, so I could clean it up and commit with the appropriate commit message if you'd like.

bbangert:
@Idem Hi! I'd be happy to help with fixing the commit if you need any assistance. A good commit message for this would look like:
```
fix: ensure every exception inherit from KazooException

Closes #383
```

You could rebase this branch against master, amend the commit, then force-push to update the PR.

bbangert:
Closing this for now, please re-open if/when you have time to update it.

PR #448, opened 2017-06-02
Use epoll when available to support fds > 1023

### LINKED ISSUE #171: kazoo patch to workaround select's file descriptor limitation
I wrote a patch to deal with the fact that select cannot handle file descriptors larger than 1024 (This problem is common to most posix select implementations).  I would like to upstream this change.

This patch is only applicable to posix based systems (Windows does not support epoll).  

patch: https://www.dropbox.com/s/9dae2bfk1jfilwr/kazoo-1.3.1.patch

original thread: https://groups.google.com/forum/#!topic/python-zk/kzX_8BBBpCM

### LINKED ISSUE #266: Use of select.select() in threading handler is a problem
I ran into a situation where an application (with a ton of open FDs) was calling my application in a popen and did not use close_fd. So the single fd passed into select.select in connection.py had a int value that was too large and select.select throws a ValueError. 

Using some other poller may be wise. 

Take a look at asyncio.selectors.DefaultSelector for a really nice solution to different poller support on different OS's

packysauce:
Here's my take on the fd > 1023 issue. I'm open to suggestions for the unit test.

Major points:
When epoll is available, and the highest fd in use is > 1023, route through epoll.
Otherwise, use the existing select() behavior so by and large nothing changes.

closes #266
closes #171

### DIFF
--- a/kazoo/handlers/threading.py
+++ b/kazoo/handlers/threading.py
@@ -18,6 +18,10 @@ import select
 import socket
 import threading
 import time
+import six
+
+from collections import defaultdict
+from itertools import chain
 
 import kazoo.python2atexit as python2atexit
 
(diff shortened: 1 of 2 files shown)


bbangert:
How does this change coverage? I'll be getting automatic coverage checking up next to make it a bit easier to track.

Otherwise I think this looks good, the commits need to be squashed and updated per the CONTRIBUTING doc style guide. Thanks!

packysauce:
Alright I think I finally figured out how to get the commits looking the way we want them.

bbangert:
The easiest way I've found is to use ``git reset COMMITBEFORENEWONE``, stash the changes, rebase to master, then stash pop to re-apply the changes, and make the single commit before force-pushing to the branch. There might be an easier way for more powerful git wizards, but that's how I usually end up doing it.

packysauce:
ended up resetting to commit^, then rebase -i and squash, then force push. Finally down to one commit though. Is github's automatic squash-merging not available for automatic merges?

bbangert:
It doesn't seem apply quite right for some reason. For the git commit message, it should look like this:
```
feat(core): use epoll when available to support fds > 1023

When epoll is available, and the highest fd in use is > 1023, route through epoll.
Otherwise, use the existing select() behavior so by and large nothing changes.

Closes #266, #171
```

packysauce:
No, it's my bad. I failed to properly grok the CONTRIBUTING.md guidelines. I'm totally ok with the "hassle" to keep things tidy :)

bbangert:
Thanks for the PR!

PR #449, opened 2017-06-03
Add missed parens to function call.

bsanders:
The second clause of the conditional (checking if the lock is held) will always evaluate to True.


bbangert:
Thanks! Not terribly reassuring that no test catches this, since it could be broken again in the future. Can you update the commit to the commit style documented in CONTRIBUTING? Otherwise I can edit, though I'd like to retain your credit, since I didn't merge your other PR that addressed this.

Just changing the commit to ``fix: add missed parens to LockingQueue function call`` would be great.

PR #450, opened 2017-06-12
fix(core): revert PR #305 SetWatches which caused RuntimeError

bbangert:
PR #305 introduced a feature to restore watches on reconnect.
Unfortunately this introduced RuntimeError's under various cases, so
reverting it is necessary.


bbangert:
For further reference, the TreeCache PR did not work with this, and the following StackOverflow question appears related:
https://stackoverflow.com/questions/44448694/kazoo-restarted-zookeeper-the-watch-threw-an-exception

PR #464, opened 2017-07-24
docs: minor tweaks to contributing.md

jeffwidman:
Fix readability, tense, grammar, etc.


CONTRIBUTING.md
```diff
@@ -27,8 +27,8 @@ your new repository and run the build scripts:
 
 You need to have some supported version of Python installed and have
 it available as `python` in your shell. To run Zookeeper you also
-need a Java runtime (JRE or JDK) version 6 or 7. To run tests, you
-need to have the tox, Python testing tool, to be installed in your shell.
+need a Java runtime (JRE or JDK) version 6, 7 or 8. To run tests, you
```
jeffwidman: Is this dependency on JDK/JRE 6 or 7 hardcoded anywhere? I added 8 now that it's in GA, but personally I'd rather remove the version numbers altogether as it's one less thing that needs updating.

bbangert: I'm not sure, the dependency is whatever version Zookeeper wants. I'd imagine we could likely just say 6+ since afaik none of the newer ones are going to break old things.

hannosch: It makes sense to replace this with something like: "To run Zookeeper you also need a Java runtime (JRE or JDK). Please refer to the Zookeeper documentation for compatible Java versions for each Zookeeper version."

jeffwidman:
Thanks for the suggested wording @hannosch, I updated the PR accordingly.

This is ready to be reviewed/merged.

PR #467, opened 2017-07-24
chore: stop bundling virtualenv

### LINKED ISSUE #462: Stop bundling virtualenv
Is there a reason this project bundles `virtualenv`? 

https://github.com/python-zk/kazoo/tree/master/sw

It is referenced in the `Makefile`: https://github.com/python-zk/kazoo/blob/83b9f94e32594516a65cd90fcf4fc16a02156467/Makefile#L31

But I don't think it's needed anymore. For tests, `tox` handles all the venv's, and for development, modern python 3 already includes the `venv` module. 

Note to self: If removing this, also remove these references to the `sw` folder:
https://github.com/python-zk/kazoo/search?utf8=%E2%9C%93&q=sw&type=

cc @harlowja / @bbangert as I think you two are most likely to know history here

jeffwidman:
We bundled our own virtualenv copy to work around bugs in travis-ci
(0cdb38186eeffe8306315f0ddf76dd003c753432). That's a 5+ year old bug
which isn't around anymore.

So no need to keep bundling virtualenv.

closes #462

@hannosch note that my `Makefile` foo isn't that great, so I copy/pasted your suggested fix. Please feel free to suggest further improvementes.


hannosch:
The test failure looks like a flaky test to me, ignoring.

PR #470, opened 2017-07-27
Drop python 2.6 compatible gevent

jeffwidman:
We dropped python 2.6 support a while ago, so no need for the added complexity here.


jeffwidman:
Oops, I totally forgot. Fixed.

PR #476, opened 2017-08-24
fix: Remove use of "async" as a variable

### LINKED ISSUE #455: Remove the use of variables named async or await
In Python 3.7 the use of `async` or `await` as variable names will cause errors as they are reserved keywords. Kazoo does this, we should get this fixed.

```
/data/users/cooper/binary#link-tree/kazoo/recipe/watchers.py:425: DeprecationWarning: 'async' and 'await' will become reserved keywords in Python 3.7
```
e.g.
https://github.com/python-zk/kazoo/blob/master/kazoo/recipe/watchers.py#L425-L427

tgockel:
The use of "async" and "await" as variable are deprecated in Python 3.7. This patch removes the use of "async" (there is no usage of "await" to fix).

Closes #455

### DIFF
--- a/kazoo/recipe/partitioner.py
+++ b/kazoo/recipe/partitioner.py
@@ -192,7 +192,7 @@ class SetPartitioner(object):
 
         # Now watch the party and set the callback on the async result
         # so we know when we're ready
-        self._child_watching(self._allocate_transition, async=True)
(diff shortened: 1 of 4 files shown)


jeffwidman:
Thank you!

PR #483, opened 2017-11-03
fix: need gevent >= 1.2, not > 1.1

jeffwidman:
Python 2.6 support was dropped in 1.2, so https://github.com/python-zk/kazoo/commit/2e8dcd3836d01640f07e8de911cdfb3639f97d20 simplified the dependency graph to require `gevent>1.1`. However, bugfix releases to the 1.1 series (1.1.1, etc) will satisfy this requirement, so we should actually be requiring `gevent >= 1.2`

### DIFF
--- a/requirements_gevent.txt
+++ b/requirements_gevent.txt
@@ -1,1 +1,1 @@
-gevent>1.1
+gevent>=1.2
--- a/setup.py
+++ b/setup.py
@@ -28,16 +28,16 @@ tests_require = install_requires + [
 
 if not (PYTHON3 or PYPY):
     tests_require += [
-        'gevent',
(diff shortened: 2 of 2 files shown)


jeffwidman:
@hannosch do you mind approving again? I merged #484 first, and then Github complained that i needed to merge that into this first, but when I solved that by simply rebasing, Github dismissed your review... The change is the exact same, simply rebased.

jeffwidman:
@hannosch sorry, do you mind reviewing one more time? I realized right after my last comment that we weren't applying this version pinning in `setup.py`, and we really ought to do it there so that no one is able to accidentally install with an unsupported gevent/eventlet version.

PR #487, opened 2017-11-06
fix(core): resolve race in IAsyncResult.wait()

### LINKED ISSUE #485: IAsyncResult wait() has a race.
The wait() method of the IAsyncResult has a race. If it is called after the asynchronous operation has completed, the Condition wait() will be stuck forever.

The attached bug.py.txt file demonstrates the problem. The key to triggering it is to call time.sleep(1) before calling wait().
[bug.py.txt](https://github.com/python-zk/kazoo/files/1442800/bug.py.txt)

Basically, the handlers.utils.AsyncResults.wait() method needs to only call the condition.wait() if the result is not ready(). I've attached a git diff output to illustrate the fix.

[diff.txt](https://github.com/python-zk/kazoo/files/1442798/diff.txt)

Thanks,

Jim Hanko

jgh-ds:
closes #485

### DIFF
--- a/kazoo/tests/test_threading_handler.py
+++ b/kazoo/tests/test_threading_handler.py
@@ -221,8 +221,8 @@ class TestThreadingAsync(unittest.TestCase):
         th.daemon = True
         th.start()
 
-        # if the wait() didn't sleep (correctly), cv will be set quickly;
-        # if it did sleep, cv will not be set yet and this will timeout
+        # if the wait() didn't sleep (correctly), cv will be set quickly
+        # if it did sleep, the cv will not be set yet and this will timeout
         cv.wait(10)
         eq_(cv.is_set(), True)
         th.join()
--- a/kazoo/tests/test_threading_handler.py
+++ b/kazoo/tests/test_threading_handler.py
@@ -220,8 +220,6 @@ class TestThreadingAsync(unittest.TestCase):
         th.start()
         cv.wait(1)
         eq_(cv.is_set(), True)
-        if not cv.is_set():
-            async_result.set("kick")
         th.join()
 
     def test_set_before_wait(self):
--- a/kazoo/tests/test_threading_handler.py
+++ b/kazoo/tests/test_threading_handler.py
@@ -225,9 +225,6 @@ class TestThreadingAsync(unittest.TestCase):
         # if it did sleep, cv will not be set yet and this will timeout
         cv.wait(5)
         eq_(cv.is_set(), True)
-        # ensure the waiter is awakened no matter what
-        if not cv.is_set():
-            async_result.set("extra-kick")
         th.join()
 
     def test_set_before_wait(self):
--- a/kazoo/tests/test_threading_handler.py
+++ b/kazoo/tests/test_threading_handler.py
@@ -210,7 +210,7 @@ class TestThreadingAsync(unittest.TestCase):
         def wait_for_val():
             try:
                 # NB: should not sleep
-                val = async_result.wait(10)
+                async_result.wait(10)
                 cv.set()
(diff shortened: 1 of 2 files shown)


jgh-ds:
I thought about adding a test, but I didn't do it because a failure would cause the test code to hang forever. However, it would be very simple to do. Should I add one?

jgh-ds:
I didn't mean to close it.

jgh-ds:
Updated jgh-ds:master with a new test case for the issue: test_wait_race()

jgh-ds:
It should be ready to merge now, including a test case.

Thanks.

kazoo/tests/test_threading_handler.py
```diff
@@ -200,6 +200,28 @@ def wait_for_val():
         eq_(lst, [True])
         th.join()
 
+    def test_wait_race(self):
```
jeffwidman: > I thought about adding a test, but I didn't do it because a failure would cause the test code to hang forever.

Oh yeah. I forgot about that. 😄 

Your test looks good to me, however it's just complicated enough that it's not immediately apparent what you're trying to test. So do you mind adding a simple docstring that explains _what_ this is trying to test/guard against?

Maybe something like:

    """Test that there is no race condition in `IAsyncResult.wait()`.
         
    Guards against https://github.com/python-zk/kazoo/issues/485 reappearing.
    """"

kazoo/tests/test_threading_handler.py
```diff
@@ -200,6 +200,28 @@ def wait_for_val():
...
+
+        cv = threading.Event()
+
+        def wait_for_val():
+            try:
+                # NB: should not sleep
+                async_result.wait(10)
+                cv.set()
+            except ImportError:
+                pass
```
jeffwidman: why is this needed?

jgh-ds: Copied from test_wait(). I'll remove it.

jgh-ds:
Added the doc string and removed the unnecessary try/except from the test case. Also verified that failing the test case will not hang the testing process.

jeffwidman:
I'm going to go ahead and merge. The test is the only part I'm slightly unsure of, but I am confident the race condition exists and your fix looks good.

jgh-ds:
I've determined that the part of the test which is:
```
        if not cv.is_set():
            async_result.set("kick")
```
is not needed (but is harmless). I was thinking that it would be necessary in the failure case, but the assertion failure from eq_(_) keeps it from being run in that case anyway. Since the thread has a timeout anyway, it cannot hang the testing process.

Would it be better to remove it?

jeffwidman:
Yes, please remove it. No one later will realize it can be removed.

Re: the merge, I accidentally clicked the wrong button, then had to wait til tests passed...

jgh-ds:
The dead code in the test is removed now.

jgh-ds:
Apparently the removed code is needed in Python 3.5 and 3.6 to avoid hanging.

jgh-ds:
The test failures are not on the new test case. The actual IAsyncResult.wait() code is the same as the first commit, where all the tests pass.

Could other test cases depend on the old (broken) behavior? The latest errors are on test_race_condition_new_partitioner_steals_the_lock

jgh-ds:
I've verified that the failing test case in test_race_condition_new_partitioner_steals_the_lock does not use the modified IAsyncResult.wait() code.
It appears that test_race_condition_new_partitioner_steals_the_lock is sensitive to execution performance (using 0.1 second and 0.2 second time constants). Perhaps the Travis environment is slower now.

jgh-ds:
It should be good to go now.

PR #489, opened 2017-12-11
Fix a batch of known issues of TreeCache recipe

tonyseek:
This pull request fixes some edge bugs which were found in our production, and improves performance by approaching the behaviors of Apache Curator:

1. The new TreeCache uses a standalone background queue instead of using the Kazoo handler's one.
2. The new TreeCache ignores unknown exceptions of async operations instead of crashing.
3. The tree refreshing operations send `\0` to a pipe or socketpair for waking up the connection routine. Those operations will block the session watcher callback and has been moved to the background queue.

Thanks for @mozillazg for detecting those issues.

### DIFF
--- a/kazoo/handlers/eventlet.py
+++ b/kazoo/handlers/eventlet.py
@@ -76,11 +76,13 @@ class SequentialEventletHandler(object):
 
     """
     name = "sequential_eventlet_handler"
+    queue_impl = green_queue.LightQueue
+    queue_empty = green_queue.Empty
 
     def __init__(self):
         """Create a :class:`SequentialEventletHandler` instance"""
-        self.callback_queue = green_queue.LightQueue()
-        self.completion_queue = green_queue.LightQueue()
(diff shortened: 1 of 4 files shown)


tonyseek:
@jeffwidman Thanks for your review. I have added regression test and commit details.

(It seems that Travis CI + ZooKeeper are not very stable)

jeffwidman:
Hmm... I clicked the "update branch" button on github assuming it would rebase, but it did a merge commit instead... I'm hesitant to merge as I'm not sure what Github will handle it under the covers and I'd rather keep a clean'ish git history... do you mind rebasing and dropping my last commit? We just had a baby, so that's why I've been slow to respond, but I should be a bit more responsive going forward...

tonyseek:
Sure. I have pushed the branch based on up-to-date master.

Congratulations on your new baby. 😄

jeffwidman:
Thanks!

PR #490, opened 2017-12-14
fix (recipe): don't set creation watch on lock predecessor node

packysauce:
Using exists will register a creation, as well as update and deletion
watch on the given node. This introduces a race, where the predecessor
node can get deleted between the call to getChildren() and exists().
If that happens, the exists() sets a create watch on a node that will
never be created, leaks a create watch.

### DIFF
--- a/kazoo/recipe/lock.py
+++ b/kazoo/recipe/lock.py
@@ -246,11 +246,14 @@ class Lock(object):
             predecessor = self.path + "/" + predecessor
             self.client.add_listener(self._watch_session)
             try:
-                if self.client.exists(predecessor, self._watch_predecessor):
-                    self.wake_event.wait(timeout)
-                    if not self.wake_event.isSet():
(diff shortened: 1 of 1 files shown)


packysauce:
I'd also really appreciate any input on how to augment the tests to catch this kind of case. We've only seen it in prod, and I can't really think of a trivial repro.

packysauce:
Hope everyone had a nice holiday, got a few moments to check out this PR?

packysauce:
I'm not sure 1) Why that pypy test failed 2) How to make it run again to check for transient problems. Anyone get a chance to review my change?

packysauce:
Alrighty, rebased and things are lookin happier

Thanks for the pro tips btw!

packysauce:
Mind giving this a once over one more time?

jeffwidman:
Thanks!

PR #493, opened 2018-02-01
client: Allow SSL use when communicating with Zookeeper, fixes #382

### LINKED ISSUE #382: Add support for SSL
As SSL is [supported by Zookeeper starting from 3.5.1](https://issues.apache.org/jira/browse/ZOOKEEPER-2125) it would be nice to have it in kazoo.

emonty:
Zookeeper 3.5 supports SSL for client communications, this commit
adds support for it on the Kazoo side.

Note that you need to give the client the key, certificate and CA
files.


Shrews:
Yes please. Much needed.

PR #495, opened 2018-02-14
fix: Pass watch as keyword arg instead of positional arg

arushiagg:
We instrument calls to async functions, but are unable to get the watch parameter because it is not being passed as part of kwargs. This will allow us to capture the watch parameter.


arushiagg:
Could someone take a look at this simple change?

jeffwidman:
Thanks!

PR #501, opened 2018-04-10
fix(recipe): Correct LockingQueue to work with multiple readers.

banburybill:
The current LockingQueue implementation implicitly assumes (in check_for_updates()) that once it has a list of available and locked queue items, the state of the first available but not taken item will not change before it can take the item. This is not true in the presence of multiple reader processes. This has two negative consequences:

1. If the item is processed and removed by another process before the item is taken, the taken entry is created and remains until the connection is closed. If connections are long-lasting, taken entries for processed and removed queue items accumulate.
2. If the item is either taken or processed and removed by another process, the call to _inner_get() will block until there is more activity on the queue, even if there are more items available for processing. Reader processes will be blocked unnecessarily.

So, remove the assumption. If a take fails, consider other items in the queue before blocking. And tidy taken entries if, after creation, the entry is found to be no longer on the queue.

### DIFF
--- a/kazoo/recipe/queue.py
+++ b/kazoo/recipe/queue.py
@@ -239,7 +239,7 @@ class LockingQueue(BaseQueue):
         if self.processing_element is None:
             return False
         lock_id, _ = self.processing_element
-        lock_path = "{path}/{id}".format(path=self._lock_path, id=lock_id)
+        lock_path = self._make_lock_path(lock_id)
         self.client.sync(lock_path)
         value, stat = self.client.retry(self.client.get, lock_path)
         return value == self.id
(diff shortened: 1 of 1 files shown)


jeffwidman:
@banburybill my apologies, I saw https://github.com/python-zk/kazoo/pull/523/files before I saw this one...

I _think_ they solve the same problem, but I'm not 100% sure... if there are still outstanding issues that your PR solves, can you rebase and then I'll take another look? Also, please make sure not only your PR, but also your commit messages follow the guidelines mentioned here: https://github.com/python-zk/kazoo/blob/master/CONTRIBUTING.md#git-commit-guidelines so that our changelog generator can parse them.

I will have more time the next few weeks for open source, so hopefully can get this handled a lot quicker than before.

jeffwidman:
Nudge @banburybill

jeffwidman:
Nudge @banburybill can you address my comment ☝️ 

I'm going to close for now, but happy to re-open if you have time to add some thoughts.

PR #509, opened 2018-06-09
fix(core): Use a copy of auth data when reconnecting

arushiagg:
It is possible to race between processing a new addAuth request(which updates the client.auth_data set) and iterating through it during reconnect. To avoid set changes during iteration, make a copy.

### DIFF
--- a/kazoo/protocol/connection.py
+++ b/kazoo/protocol/connection.py
@@ -1,6 +1,7 @@
 """Zookeeper Protocol Connection Handler"""
 from binascii import hexlify
 from contextlib import contextmanager
+import copy
 import logging
 import random
 import select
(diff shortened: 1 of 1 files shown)


arushiagg:
One of the tests seems to have had a connection loss error. Does not seem related to the change.

kazoo/protocol/connection.py
```diff
@@ -646,7 +647,10 @@ def _connect(self, host, port):
             client._session_callback(KeeperState.CONNECTED)
             self._ro_mode = None
 
-        for scheme, auth in client.auth_data:
+        # Get a copy of the auth data before iterating, incase it is
```
jeffwidman: typo: `incase` --> `in case`

arushiagg:
Rebased, and updated to fix commit message/typo.

PR #512, opened 2018-07-04
feat(core): Added SASL DIGEST-MD5 support

StephenSorriaux:
This PR adds support for **DIGEST-MD5** SASL authentification to Zookeeper.

Usage is as same as before except that we can now use the 'sasl' mechanism:
```
zk = KazooClient(hosts='172.20.0.2:2181', auth_data=[('sasl', 'user:password')])
zk.start()
...
```
I mainly followed [Zookeeper's client-server mutual authentication](https://cwiki.apache.org/confluence/display/ZOOKEEPER/Client-Server+mutual+authentication) documentation for this implementation.

It uses the [pure-sasl library](https://github.com/thobbs/pure-sasl) to manage the SASL challenge. It is a very well documented library to use for SASL authentification and can also be used if we want to add a **GSSAPI** support for Zookeeper.

Some tests also have been added.

Thank you

Stephen

### DIFF
--- a/kazoo/protocol/connection.py
+++ b/kazoo/protocol/connection.py
@@ -25,6 +25,7 @@ from kazoo.protocol.serialization import (
     Ping,
     PingInstance,
     ReplyHeader,
+    SASL,
     Transaction,
     Watch,
     int_struct
@@ -39,6 +40,11 @@ from kazoo.retry import (
     ForceRetryError,
     RetryFailedError
 )
+try:
+    from puresasl.client import SASLClient
+    PURESASL_AVAILABLE = True
+except ImportError:
+    PURESASL_AVAILABLE = False
 
 
 log = logging.getLogger(__name__)
@@ -153,6 +159,8 @@ class ConnectionHandler(object):
         self._ro_mode = False
 
         self._connection_routine = None
+
+        self.sasl_cli = None
 
     # This is instance specific to avoid odd thread bug issues in Python
     # during shutdown global cleanup
@@ -416,6 +424,24 @@ class ConnectionHandler(object):
                 async_object.set(True)
         elif header.xid == WATCH_XID:
             self._read_watch_event(buffer, offset)
+        elif self.sasl_cli and not self.sasl_cli.complete:
+            # SASL authentication is not yet finished, this can only
+            # be a SASL packet
+            self.logger.log(BLATHER, 'Received SASL')
+            try:
+                challenge, _ = SASL.deserialize(buffer, offset)
+            except Exception:
+                raise ConnectionDropped('error while SASL authentication.')
+            response = self.sasl_cli.process(challenge)
+            if response:
+                # authentication not yet finished, answering the challenge
+                self._send_sasl_request(challenge=response,
+                                        timeout=client._session_timeout)
+            else:
+                # authentication is ok, state is CONNECTED
+                # remove sensible information from the object
+                client._session_callback(KeeperState.CONNECTED)
+                self.sasl_cli.dispose()
         else:
             self.logger.log(BLATHER, 'Reading for header %r', header)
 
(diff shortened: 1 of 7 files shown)


kazoo/protocol/connection.py
```diff
@@ -1,12 +1,14 @@
 """Zookeeper Protocol Connection Handler"""
 from binascii import hexlify
 from contextlib import contextmanager
+
```
jeffwidman: why the new line? `stdlib` imports should be grouped together per `pep8`

kazoo/protocol/connection.py
```diff
@@ -1,12 +1,14 @@
...
 import logging
 import random
 import select
 import socket
 import sys
 import time
+from puresasl.client import SASLClient
```
jeffwidman: Since this is a 3rd-party lib, it should be an optional dependency (which you correctly handled in the `setup.py`)... So shouldn't this be wrapped in a `try`/`except` to handle the `ImportError` when this library is not available?

StephenSorriaux:
Thank you!

I will update all my PRs next week.

StephenSorriaux:
@jeffwidman I just updated this PR following your advises. In case pure-sasl library is missing, the client will fall back connection to zk without any auth. Example:
```
>>> zk.start()
2018-10-01 12:02:14,215 - kazoo.client - INFO - Connecting to 172.19.0.2:2181
2018-10-01 12:02:14,216 - kazoo.client - DEBUG - Sending request(xid=None): Connect(protocol_version=0, last_zxid_seen=0, time_out=10000, session_id=0, passwd='\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00', read_only=None)
2018-10-01 12:02:14,752 - kazoo.client - WARNING - Pure-sasl library is missing while sasl authentification is configured. Please install pure-sasl library to connect using sasl. Now falling back connecting WITHOUT any authentification.
2018-10-01 12:02:14,752 - kazoo.client - INFO - Zookeeper connection established, state: CONNECTED

--- Zookeeper side
2018-10-01 10:02:14,217 [myid:] - INFO  [NIOServerCxn.Factory:0.0.0.0/0.0.0.0:2181:NIOServerCnxnFactory@215] - Accepted socket connection from /172.19.0.1:44144
2018-10-01 10:02:14,218 [myid:] - INFO  [NIOServerCxn.Factory:0.0.0.0/0.0.0.0:2181:ZooKeeperServer@948] - Client attempting to establish new session at /172.19.0.1:44144
2018-10-01 10:02:14,752 [myid:] - INFO  [SyncThread:0:ZooKeeperServer@693] - Established session 0x10000619e220005 with negotiated timeout 10000 for client /172.19.0.1:44144
```

kazoo/protocol/connection.py
```diff
@@ -660,13 +686,53 @@ def _connect(self, host, port):
...
         else:
-            client._session_callback(KeeperState.CONNECTED)
             self._ro_mode = None
-
-        for scheme, auth in client.auth_data:
-            ap = Auth(0, scheme, auth)
-            zxid = self._invoke(connect_timeout / 1000.0, ap, xid=AUTH_XID)
-            if zxid:
-                client.last_zxid = zxid
+            if client.use_sasl and self.sasl_cli is None:
```
jeffwidman: @StephenSorriaux I just realized you deleted the blank line and indented the following code so that it's under the `else` in 688... why?

Previously it would run regardless of the code in lines 685-689, but now it's only if you hit the `else` clause...

You have paid attention to details in your PR, so my guess is that I'm just not understanding something, so can you clarify?

StephenSorriaux: @jeffwidman You're right, I clearly overlooked this part when adding the behaviour handling the case where pure-sasl's lib is absent. There is no reason not doing auth on RO connection... I'm sorry, I will submit a PR tomorow to fix that

PR #494, opened 2018-02-01
client: Allow SSL use when communicating with Zookeeper, fixes #382

emonty:
Zookeeper 3.5 supports SSL for client communications, this commit
adds support for it on the Kazoo side.

Note that you need to give the client the key, certificate and CA
files.

Co-Authored-By: Monty Taylor <mordred@inaugust.com>

PR #513, opened 2018-07-06
feat(core): Added SSL support

StephenSorriaux:
This PR adds SSL support for Zookeeper connection. I took back the commit from https://github.com/python-zk/kazoo/pull/494

Usage is:
```
zk = KazooClient(hosts='172.20.0.2:2281', use_ssl=True, keyfile='/path/to/key.pem', certfile='/path/to/cert.pem', ca='/path/to/ca.pem', keyfile_password='password')
zk.start()
...
```
When playing with self signed certificate, it is possible to bypass certs verification using the paramater 'verify_certs':
```
zk = KazooClient(hosts='172.20.0.2:2281', use_ssl=True, keyfile='/path/to/key.pem', certfile='/path/to/cert.pem', keyfile_password='password', verify_certs=False)
zk.start()
...
```
I am not sure on how to add tests since I need to add a specific SSL configuration for only these tests and that I may need to dynamically generate certificates and keystore / truststore. I would be grateful if I can get some hints on how to add those tests.

Stephen

### DIFF
--- a/kazoo/handlers/utils.py
+++ b/kazoo/handlers/utils.py
@@ -187,7 +187,8 @@ def create_tcp_socket(module):
 
 def create_tcp_connection(module, address, timeout=None,
                           use_ssl=False, ca=None, certfile=None,
-                          keyfile=None, keyfile_password=None):
+                          keyfile=None, keyfile_password=None,
+                          verify_certs=True):
     end = None
     if timeout is None:
         # thanks to create_connection() developers for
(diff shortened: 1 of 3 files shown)


kazoo/handlers/utils.py
```diff
@@ -194,17 +199,48 @@ def create_tcp_connection(module, address, timeout=None):
...
+                context.load_verify_locations(ca)
+            if certfile and keyfile:
+                context.verify_mode = (
+                    ssl.CERT_REQUIRED if verify_certs else ssl.CERT_NONE
+                )
+                context.load_cert_chain(certfile=certfile,
+                                        keyfile=keyfile,
+                                        password=keyfile_password)
+            try:
+                # Query the address to get back it's address family
```
jeffwidman: no `'` in this case

jeffwidman:
Fixes #382

PR #518, opened 2018-07-25
fix(core): negative timeouts should not be generated

brycedrennan:
Encountered a strange situation where `timeout_at` was set to a negative value.  This caused python sock.settimeout() to throw ValueError('Timeout value out of range',)

The probably-overly-broad `try...except` below couldn't handle the ValueError that was thrown as a result. Thus a TypeError("'ValueError' object is not subscriptable") was thrown.

Dump of variables at the moment the when exception occurred:

```
timeout_at	-6.888538122177124
sock	    None
end	        1532524647.0125406
timeout     10.0
address	    ('10.0.100.82', 2181)
module	    <module 'socket' from '/usr/local/lib/python3.6/socket.py'>
```

This would seem to imply that somehow ~16.8 seconds elapsed between `end` getting set and timeout_at being calculated. Probably a weird edge case.

One other thing, reading the python socket settimeout docs, its wasn't clear to me if the behavior of settimeout(0.0) was to immediately timeout or to have no timeout.  Thus I set it to an _almost_ immediate timeout value.

### DIFF
--- a/kazoo/handlers/utils.py
+++ b/kazoo/handlers/utils.py
@@ -199,7 +199,11 @@ def create_tcp_connection(module, address, timeout=None,
     sock = None
 
     while end is None or time.time() < end:
-        timeout_at = end if end is None else end - time.time()
+        # max() is used to handle a weird edge case where timeout_at
+        # was being set to a negative number. 0.0001 was arbitrarily
+        # picked as a small non-zero number. more info here:
+        # https://github.com/python-zk/kazoo/pull/518
+        timeout_at = end if end is None else max(end - time.time(), 0.0001)
         if use_ssl:
             # Disallow use of SSLv2 and V3 (meaning we require TLSv1.0+)
             context = ssl.SSLContext(ssl.PROTOCOL_SSLv23)


jeffwidman:
This looks excellent, and I'd like to merge. 

Can you amend the inline comment that explains this line of code to also clarify that `max()` is used to handle edge cases where unusual time shift results in negative timeouts?

Also, would you be willing to chase down the behavior of `settimeout(0.0)`? If you're not comfortable reading the Python source, a quick post on StackOverflow or Reddit should provide an answer... If you still need the `0.0001`, it'd be good to have an inline comment explaining the purpose as otherwise it's confusing why that number was chosen. I know it's in git history, but I've seen that munged before, so prefer to have this one annotated in the code.

If we can get these two things handled, I'll happily merge

brycedrennan:
Thanks @jeffwidman , I'll add an explanatory comment.  

With regards to settimeout, I think the documentation is probably clear enough.  I'm just not familiar enough with sockets to confidently interpret the documentation. Here are the relevant explanations from the python documentation. Maybe you or someone else can confidently say 0 is okay given the explanation below?

> If a non-zero value is given, subsequent socket operations will raise a timeout exception if the timeout period value has elapsed before the operation has completed. If zero is given, the socket is put in non-blocking mode. If None is given, the socket is put in blocking mode. ([src](https://docs.python.org/3/library/socket.html#socket.socket.settimeout))

> A socket object can be in one of three modes: blocking, non-blocking, or timeout. Sockets are by default always created in blocking mode, but this can be changed by calling setdefaulttimeout().
> In blocking mode, operations block until complete or the system returns an error (such as connection timed out).
> In non-blocking mode, operations fail (with an error that is unfortunately system-dependent) if they cannot be completed immediately: functions from the select can be used to know when and whether a socket is available for reading or writing.
> In timeout mode, operations fail if they cannot be completed within the timeout specified for the socket (they raise a timeout exception) or if the system returns an error. ([src](https://docs.python.org/3/library/socket.html#socket-timeouts))

brycedrennan:
@jeffwidman Ready for your review again.

I added a comment as well as rebased to address the changes that have occurred in master.

jeffwidman:
Sorry, I haven't had time yet to investigate the `settimeout` bit, I hope to do so later this week or next...

StephenSorriaux:
AFAICT we don't want to put the socket in non-blocking mode since it would change the way we are interacting with it.
Since the problem is caused by the calculation of `end - time.time()` when the process froze before, shouldn't replacing `end - time.time()` with `timeout` be the best way to fix it? It seems to me that we only want to define the timeout value for the socket here, which has already been defined before.

StephenSorriaux:
@brycedrennan Thank you for your PR but I will close it since #534 adressed the same issue and was merged.

PR #521, opened 2018-09-10
KazooRetry: Do uniform jitter over the whole backoff interval

andreasgudmundsson:
The previous implementation would add a fixed amount of jitter around
the calculated back-off time. Retry attempts were thus clustered
around the exponentially spaced backoff points.

This patch does exponential backoff but uniformly spreads the retries
over an interval [0, backoff**attempt]

The following graphs show the connection rate/s when trying to open 20k new connections
to ZK. Each client was configured with, backoff=2, delay=10 and max_delay=180. The ZK in question is rate-limited and that explains the ceiling you see on the second graph. 

**Old Jitter**
The connections attempts come in herds after the current delay interval has passed.
After over 30mins only about 10k connections had been established and that is where I cut-off the graph.

zk_connections_no_jitter

**Uniform Jitter**
The connection attempts reach the rate-limit then after 8 minutes the connection rate
falls below the limit and the remaining clients connect over a period of 5 mins. The whole
process takes 13 minutes.

zk_connections_patched

### DIFF
--- a/kazoo/client.py
+++ b/kazoo/client.py
@@ -79,7 +79,6 @@ _RETRY_COMPAT_DEFAULTS = dict(
     max_retries=None,
     retry_delay=0.1,
     retry_backoff=2,
-    retry_jitter=0.8,
     retry_max_delay=3600,
 )
 
@@ -87,7 +86,6 @@ _RETRY_COMPAT_MAPPING = dict(
     max_retries='max_tries',
     retry_delay='delay',
     retry_backoff='backoff',
-    retry_jitter='max_jitter',
     retry_max_delay='max_delay',
 )
 
--- a/kazoo/retry.py
+++ b/kazoo/retry.py
@@ -42,19 +42,17 @@ class KazooRetry(object):
         SessionExpiredError,
     )
 
-    def __init__(self, max_tries=1, delay=0.1, backoff=2, max_jitter=0.8,
+    def __init__(self, max_tries=1, delay=0.1, backoff=2,
                  max_delay=60, ignore_expire=True, sleep_func=time.sleep,
                  deadline=None, interrupt=None):
         """Create a :class:`KazooRetry` instance for retrying function
-        calls
+        calls with uniform jitter
 
         :param max_tries: How many times to retry the command. -1 means
                           infinite tries.
         :param delay: Initial delay between retry attempts.
         :param backoff: Backoff multiplier between retry attempts.
                         Defaults to 2 for exponential backoff.
-        :param max_jitter: Additional max jitter period to wait between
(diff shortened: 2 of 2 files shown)


kazoo/retry.py
```diff
@@ -129,8 +125,7 @@ def __call__(self, func, *args, **kwargs):
                 if self._attempts == self.max_tries:
                     raise RetryFailedError("Too many retry attempts")
                 self._attempts += 1
-                sleeptime = self._cur_delay + (
-                    random.randint(0, self.max_jitter) / 100.0)
+                sleeptime = random.randint(0, 1+int(self._cur_delay))
```
jeffwidman: style: missing spaces around the `+`

kazoo/client.py
```diff
@@ -79,15 +79,13 @@
...
     retry_backoff=2,
-    retry_jitter=0.8,
     retry_max_delay=3600,
 )
 
 _RETRY_COMPAT_MAPPING = dict(
     max_retries='max_tries',
     retry_delay='delay',
     retry_backoff='backoff',
-    retry_jitter='max_jitter',
```
jeffwidman: Note: Technically this is a backwards-incompatible change as it will stop warning that this is a deprecated param. However, it looks like this has been deprecated for over 5 years since version 1.2: https://github.com/python-zk/kazoo/commit/3d577e4f4167d72e0e1553837b602ac2da2848d0#diff-7d5ec76d88673978d0b9b679863f349dR228

So I'm in agreement with simply removing this altogether.

andreasgudmundsson:
I've changed the commit message so it should pass the commit message guidelines.

kazoo/retry.py
andreasgudmundsson: I pushed a style fix along with an updated commit message.

jeffwidman:
Awesome, merged! Thank you @andreasgudmundsson again for demonstrating with those graphs

andreasgudmundsson:
Thanks for merging the request @jeffwidman. Do you have plans for making a new release?

jeffwidman:
> Do you have plans for making a new release?

There's a couple of submitted PR's that I'm waiting on the owners to rebase/address minor review points and it'd be nice to land those first. So I'll wait a week, and then I'm traveling the following week for work, so probably the week after (3 weeks from now) I'd be willing to cut a release because that way if there's a bug in one of the recently submitted PRs, I'll have the time to merge a fix and cut a second bugfix release. If I forget, please ping me on here.

andreasgudmundsson:
> If I forget, please ping me on here.

Ping

StephenSorriaux:
Hi @andreasgudmundsson , the 2.6.0 release of Kazoo is now live!

PR #524, opened 2018-10-01
fix(recipe): No more memory leak once TreeCache was closed

tonyseek:
The current implementation of TreeCache recipe could not be closed completely because some unexpected references:

1. The registered znode event watchers hold the reference of instancemethod which are bound with `TreeNode` instances
2. The background threads in connection handler keep last reference when they are idle

This pull request fixes those memory leak problems, and verifies that by using `objgraph` in test.

We could guarantee all closed `TreeCache` will be eventually released now. (not immediately since the cycle reference between `TreeNode` and `TreeCache` must be resolved by mark-swap GC of Python)

### DIFF
--- a/kazoo/handlers/gevent.py
+++ b/kazoo/handlers/gevent.py
@@ -69,14 +69,17 @@ class SequentialGeventHandler(object):
             while True:
                 try:
                     func = queue.get()
-                    if func is _STOP:
-                        break
-                    func()
+                    try:
+                        if func is _STOP:
+                            break
+                        func()
+                    except Exception as exc:
+                        log.warning("Exception in worker greenlet")
+                        log.exception(exc)
+                    finally:
+                        del func  # release before possible idle
                 except self.queue_empty:
                     continue
-                except Exception as exc:
-                    log.warning("Exception in worker greenlet")
-                    log.exception(exc)
         return gevent.spawn(greenlet_worker)
 
     def start(self):
--- a/kazoo/tests/test_cache.py
+++ b/kazoo/tests/test_cache.py
@@ -1,17 +1,44 @@
 import gc
+import importlib
 import uuid
 
 from mock import patch, call, Mock
 from nose.tools import eq_, ok_, assert_not_equal, raises
 from objgraph import count as count_refs_by_type
 
-from kazoo.testing import KazooTestCase
+from kazoo.testing import KazooTestHarness
 from kazoo.exceptions import KazooException
 from kazoo.recipe.cache import TreeCache, TreeNode, TreeEvent
 
 
-class KazooTreeCacheTests(KazooTestCase):
+class KazooAdaptiveHandlerTestCase(KazooTestHarness):
+    HANDLERS = (
+        ('kazoo.handlers.gevent', 'SequentialGeventHandler'),
+        ('kazoo.handlers.eventlet', 'SequentialEventletHandler'),
+        ('kazoo.handlers.threading', 'SequentialThreadingHandler'),
+    )
 
+    def setUp(self):
+        self.handler = self.choose_an_installed_handler()
+        self.setup_zookeeper(handler=self.handler)
+
+    def tearDown(self):
+        self.handler = None
+        self.teardown_zookeeper()
+
+    def choose_an_installed_handler(self):
+        for handler_module, handler_class in self.HANDLERS:
+            try:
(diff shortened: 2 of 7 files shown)


tonyseek:
😸 It is ready for review now.

tonyseek:
Hi @jeffwidman, Thanks for you review. The commits have been revised now.

kazoo/handlers/gevent.py
```diff
@@ -69,14 +69,17 @@ def greenlet_worker():
...
-                    func()
+                    try:
+                        if func is _STOP:
+                            break
+                        func()
+                    except Exception as exc:
+                        log.warning("Exception in worker greenlet")
+                        log.exception(exc)
+                    finally:
+                        del func  # release before possible idle
```
tonyseek: > Given the attention to detail in your PR's, I suspect I'm missing something... Is there some reason that it's better to set it to None?

I had set the `func` to `None` for preventing from `NameError` if an exception was raised earlier than `func` was assigned. But it is confusing and `del func` is more explicit that it, I agree with you.

I use the nested `try-except` like the implementation of `SequentialThreadingHandler` now. The `NameError` will never occur here.

tonyseek:
I will keep the branch up-to-date for latest master. The new pushed commits are just for rebasing.

StephenSorriaux:
Hi @tonyseek, thanks for your PR. I will try to give it a review during the week.

kazoo/tests/test_cache.py
```diff
@@ -51,6 +92,27 @@ def spy_client(self, method_name):
         method = getattr(self.client, method_name)
         return patch.object(self.client, method_name, wraps=method)
 
+    def _gc_wait(self):
+        if self.handler.name == 'sequential_threading_handler':
+            while not self.client.handler.completion_queue.empty():
+                self.client.handler.sleep_func(0.1)
+        else:
+            # Let hub of gevent/eventlet be in a short-time loop
```
StephenSorriaux: Should we check here that the queue is empty?

kazoo/tests/test_cache.py
```diff
@@ -1,15 +1,53 @@
...
+    def choose_an_installed_handler(self):
+        for handler_module, handler_class in self.HANDLERS:
+            try:
+                mod = importlib.import_module(handler_module)
+                cls = getattr(mod, handler_class)
+            except ImportError:
+                continue
+            except Exception:
+                logger.warning(
+                    'The handler module %s is imported but could not be used',
```
StephenSorriaux: Did you notice cases where this can happen?

tonyseek: In my cases it only happen when the gevent installation is broken. I should delete it and make it fail fast.

tonyseek: Fixed in 6c0920c

kazoo/tests/test_cache.py
tonyseek: Modified in 6c0920c

kazoo/tests/test_cache.py
```diff
@@ -1,15 +1,48 @@
...
+from objgraph import count as count_refs_by_type
 
-from kazoo.testing import KazooTestCase
+from kazoo.testing import KazooTestHarness
 from kazoo.exceptions import KazooException
 from kazoo.recipe.cache import TreeCache, TreeNode, TreeEvent
 
 
-class KazooTreeCacheTests(KazooTestCase):
+logger = logging.getLogger(__name__)
```
StephenSorriaux: Seems like you can remove this object now since it is not used anymore

kazoo/tests/test_cache.py
```diff
@@ -1,15 +1,48 @@
+import gc
+import importlib
+import logging
```
StephenSorriaux: This import is now useless

StephenSorriaux:
Thanks again for this great PR, it is impressive. I just added 2 small comments... I think we will be ready to merge this PR asap.

tonyseek:
@StephenSorriaux My pleasure. And thank you for your review.

The unused logger has been removed in 4a26c99 now.

StephenSorriaux:
@tonyseek merged to master. Thanks again for your hard work.

jeffwidman:
Great job pushing this through @StephenSorriaux  / @tonyseek

PR #528, opened 2018-10-08
docs: make badges link to travis and pypi

nakulpathak3:
Currently the badges just open a static image of the badges. This makes it so that it opens the actual links to latest travis build and PyPi instead.

[Updated README](https://github.com/python-zk/kazoo/blob/0dfa2670393c6878896bab630825664704c9efe8/README.md).

For context: https://docs.travis-ci.com/user/status-images/ this is where I got the Travis link button command. 
And https://badge.fury.io/for/py/kazoo for the PyPi badge.

**Update**: Moved to using shields.io like it was before instead of badge.fury. Here's the [README](https://github.com/python-zk/kazoo/blob/c79a152668e166321d849aa3ed749264f3eb1b29/README.md) with shields.io.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,9 +1,9 @@
 Kazoo
 =====
 
-![Travis Build](https://travis-ci.org/python-zk/kazoo.svg?branch=master)
+[![Build Status](https://travis-ci.org/python-zk/kazoo.svg?branch=master)](https://travis-ci.org/python-zk/kazoo)
 
-![Latest Version](https://img.shields.io/pypi/v/kazoo.svg)
+[![PyPI version](https://badge.fury.io/py/kazoo.svg)](https://badge.fury.io/py/kazoo)
 
 `kazoo` implements a higher level API to [Apache
 Zookeeper](http://zookeeper.apache.org/) for Python clients.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,9 +1,9 @@
 Kazoo
 =====
 
-![Travis Build](https://travis-ci.org/python-zk/kazoo.svg?branch=master)
+[![Build Status](https://travis-ci.org/python-zk/kazoo.svg?branch=master)](https://travis-ci.org/python-zk/kazoo)
 
-![Latest Version](https://img.shields.io/pypi/v/kazoo.svg)
+[![PyPI version](https://badge.fury.io/py/kazoo.svg)](https://badge.fury.io/py/kazoo)
```
jeffwidman: I don't understand this last one:

1. Why switch away from `shields.io` as the image source?
2. Why is the URL not pointing directly to PyPI? https://pypi.org/project/kazoo/

nakulpathak3: 1. Didn't realize that it was a badge service (thought it was maybe just a static image host or something) until you mentioned it.
2. `badge.fury.io` automatically redirects to PyPi kazoo. It's just the default way to use them as mentioned in their docs.

I've added a commit to switch back to shields.io and just fix the link instead after reading your comment though.

PR #532, opened 2018-10-22
Add "Reviewed by Hound" badge

salbertson:
screenshot 2018-10-22 08 27 03

It looks like there are duplicate status updates for Hound. The capitalized one is the correct update. You should be able to [turn off branch protection](https://github.com/python-zk/kazoo/settings/branches) for `hound` to fix that.


StephenSorriaux:
Thank you, it looks nice!

I personnaly don't have the right to change branches' configuration, I just told the team to check it.

PR #534, opened 2018-10-30
fix: ensure timeout argument is nonnegative

cdonati:
Previously, a gap between calls to `time.time()` could lead to a situation where the current time was less than `end` during the `while` condition, but it was greater than `end` when assigning a value to `timeout_at`.

This is the same issue describe in https://github.com/python-zk/kazoo/pull/518, but the fix is slightly different. It avoids a redundant call to `time.time()`. I've also included a unit test that exhibits the problem when run without the fix.

### DIFF
--- a/kazoo/tests/test_utils.py
+++ b/kazoo/tests/test_utils.py
@@ -1,0 +1,78 @@
+import unittest
+
+from mock import patch
+from nose import SkipTest
+
+try:
+    from kazoo.handlers.eventlet import green_socket as socket
+    EVENTLET_HANDLER_AVAILABLE = True
+except ImportError:
+    EVENTLET_HANDLER_AVAILABLE = False
+
+
+class TestCreateTCPConnection(unittest.TestCase):
+    def test_timeout_arg(self):
+        from kazoo.handlers import utils
+        from kazoo.handlers.utils import create_tcp_connection, socket, time
+
+        with patch.object(socket, 'create_connection') as create_connection:
+            with patch.object(utils, '_set_default_tcpsock_options'):
+                # Ensure a gap between calls to time.time() does not result in
+                # create_connection being called with a negative timeout
+                # argument.
+                with patch.object(time, 'time', side_effect=range(10)):
+                    create_tcp_connection(socket, ('127.0.0.1', 2181),
+                                          timeout=1.5)
+
+                for call_args in create_connection.call_args_list:
+                    timeout = call_args[0][1]
+                    assert timeout >= 0, 'socket timeout must be nonnegative'
+
+    def test_timeout_arg_eventlet(self):
+        if not EVENTLET_HANDLER_AVAILABLE:
+            raise SkipTest('eventlet handler not available.')
+
+        from kazoo.handlers import utils
+        from kazoo.handlers.utils import create_tcp_connection, time
+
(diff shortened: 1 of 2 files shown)


kazoo/tests/test_utils.py
```diff
@@ -0,0 +1,22 @@
+import unittest
+
+from mock import patch
+
+
+class TestCreateTCPConnection(unittest.TestCase):
+    def test_timeout_arg(self):
```
StephenSorriaux: It seems to me that this test is currently calling the `create_connection` method with a timeout of `0.5`. Should we add a test to verify that when using `create_tcp_connection(socket, ('127.0.0.1', 2181), timeout=0.5)` (that means, `timeout_at` is < 0), a `socket.error` is raised? Maybe that would also be better to change this `socket.error` to something more specific.

cdonati:
@StephenSorriaux Thanks for the review. Originally, I wasn't sure if (core) was applicable here for the scope, but that makes sense now.

In addition to the tests you suggested, I also added a couple tests for nonpositive timeout values.

Let me know if the amended commits address all of your comments.

kazoo/tests/test_utils.py
```diff
@@ -0,0 +1,81 @@
+import unittest
+
+from mock import patch
+from nose import SkipTest
+
+try:
+    import eventlet
```
StephenSorriaux:
```suggestion
```

kazoo/tests/test_utils.py
```diff
@@ -0,0 +1,81 @@
...
+    from eventlet.green import threading
```
StephenSorriaux:
```suggestion
```

kazoo/tests/test_utils.py
```diff
@@ -0,0 +1,81 @@
...
+    from kazoo.handlers import eventlet as eventlet_handler
```
StephenSorriaux:
```suggestion
    from kazoo.handlers.eventlet import green_socket as socket
```

kazoo/tests/test_utils.py
```diff
@@ -0,0 +1,81 @@
...
+                for call_args in create_connection.call_args_list:
+                    timeout = call_args[0][1]
+                    assert timeout >= 0, 'socket timeout must be nonnegative'
+
+    def test_timeout_arg_eventlet(self):
+        if not EVENTLET_HANDLER_AVAILABLE:
+            raise SkipTest('eventlet handler not available.')
+
+        from kazoo.handlers import utils
+        from kazoo.handlers.eventlet import green_socket as socket
```
StephenSorriaux:
```suggestion
```

cdonati:
@StephenSorriaux Your changes look good to me. Would you prefer I use the "Apply suggestion" feature on all of them or should I push the changes you suggested as a single additional commit?

StephenSorriaux:
@cdonati I think it is better that you push the changes, I'm not sure how many commits will be created using the "Apply suggestion" feature.

cdonati:
@StephenSorriaux I just saw your comment about a single commit. I squashed the commits accordingly.

StephenSorriaux:
@cdonati Thanks again for your PR, I'm merging it to master.

PR #540, opened 2018-11-22
Reduce timeout for the first Connect() request

ralt:
In the case of a zookeeper server under pressure, it will typically try to maintain the quorum rather than handling client requests. In this kind of case, the quorum is maintained, the connection works, but the client is frozen there.

Retrying after a shorter timeout means we can reconnect to another server before losing the session altogether.

### DIFF
--- a/kazoo/protocol/connection.py
+++ b/kazoo/protocol/connection.py
@@ -660,7 +660,7 @@ class ConnectionHandler(object):
                           client.read_only)
 
         connect_result, zxid = self._invoke(
-            client._session_timeout / 1000.0, connect)
+            client._session_timeout / 1000.0 / len(client.hosts), connect)
 
         if connect_result.time_out <= 0:
             raise SessionExpiredError("Session has expired")


ralt:
Hi,

This was defined in a similar manner as `connect_timeout`. The idea being that if you have one server, then the timeout can be the full `timeout` (`/ 1`). If you have more, then it makes sense to split that, so that the client gets a chance to connect to other servers if the `Connect()` request fails.

StephenSorriaux:
You are right, this makes sense.

It is also available in the java client (https://github.com/apache/zookeeper/blob/master/zookeeper-server/src/main/java/org/apache/zookeeper/ClientCnxn.java#L1383).

Can you please amend your commit in order to follow the [CONTRIBUTING.md](https://raw.githubusercontent.com/python-zk/kazoo/master/CONTRIBUTING.md) guidelines?

ralt:
@StephenSorriaux does it look good now?

StephenSorriaux:
@ralt That is great! Thanks again for your PR.

ralt:
Thanks for the merge! :+1:

PR #543, opened 2018-12-04
fix(recipe):No more memory leak when ChildrenWatch was stopped

kulallwang:
Remove session watcher from the listener when ChildrenWatch was stopped

### DIFF
--- a/kazoo/tests/test_watchers.py
+++ b/kazoo/tests/test_watchers.py
@@ -432,7 +432,8 @@ class KazooChildrenWatcherTests(KazooTestCase):
             if fail_through:
                 return False
 
-        session_watcher = self.client.ChildrenWatch(self.path, changed)._session_watcher
+        childrenWatch = self.client.ChildrenWatch(self.path, changed)
+        session_watcher = childrenWatch._session_watcher
 
         update.wait(10)
         eq_(session_watcher in self.client.state_listeners, True)
--- a/kazoo/tests/test_watchers.py
+++ b/kazoo/tests/test_watchers.py
@@ -418,7 +418,7 @@ class KazooChildrenWatcherTests(KazooTestCase):
         update.wait(0.5)
         eq_(all_children, ['smith'])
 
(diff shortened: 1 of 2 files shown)


kazoo/tests/test_watchers.py
```diff
@@ -418,6 +418,40 @@ def changed(children):
         update.wait(0.5)
         eq_(all_children, ['smith'])
 
+    def test_remove_session_watcher_on_func_stops(self):
```
StephenSorriaux: Since this test is concerning the ChildrenWatch watcher, maybe it is better to precise it in the test's name

kazoo/tests/test_watchers.py
```diff
@@ -418,6 +418,40 @@ def changed(children):
...
+
+        def changed(children):
+            while all_children:
+                all_children.pop()
+            all_children.extend(children)
+            update.set()
+            if fail_through:
+                return False
+
+        childrenWatch = self.client.ChildrenWatch(self.path, changed)
```
StephenSorriaux: Maybe it's better to keep a snake case notation for this variable?

kazoo/tests/test_watchers.py
```diff
@@ -418,6 +418,40 @@ def changed(children):
...
+
+        fail_through.append(True)
+        self.client.create(self.path + '/' + 'smith')
+        update.wait(10)
+        eq_(session_watcher not in self.client.state_listeners, True)
+        eq_(all_children, ['smith'])
+        update.clear()
+
+        self.client.create(self.path + '/' + 'george')
+        update.wait(0.5)
```
StephenSorriaux: Isn't this value a little too small for a timeout?

kazoo/tests/test_watchers.py
kulallwang: Thank you for your comment. I use java in most time and carelessly write the variable with camel case .  I will turn it into snake case notation.

StephenSorriaux:
@kulallwang thanks again for your PR, this is now on master.

Issue #544, opened 2018-12-07
Upgrade to Kazoo 2.6.0 breaks backward compatibility with KazooRetry


ceache:
The removal of `max_jitter` from `KazooRetry` parameters in 60366d2c7910fc833991fad8e04bbe33817c0544 break upgrade from previous versions.

If this is expected, I think it should be documented.
Otherwise a simple deprecation warning and keeping the argument in an inactive form could also be done.

StephenSorriaux:
Hi,

You're right we should have kept the argument while displaying a deprecation warning message... If you have some times to fix this, please feel free to submit a PR.

jeffwidman:
@StephenSorriaux did you see my comment in https://github.com/python-zk/kazoo/pull/521#discussion_r220348402?

StephenSorriaux:
@jeffwidman No, I did not... I might be missing something but isn't the deprecation about the `KazooClient` object?

ceache:
We did not either.

We hit this because we were creating `KazooRetry` objects directly, not through the `KazooClient` constructor, which is *not* displaying any deprecation warnings.

Shouldn't the this machinery be moved to `kazoo.retry` instead and consolidated?

StephenSorriaux:
I'm closing this issue since #545 fixes it.

PR #545, opened 2018-12-11
fix(regression): Support deprecated KazooRetry argument

ceache:
Accept kazoo<=2.5.0 KazooRetry 'max_jitter' argument and display a
warning for backward compatibility.

See #544


StephenSorriaux:
@ceache thank you for your PR, this is now on master.

PR #533, opened 2018-10-23
feat(core): add support for GSSAPI authentication (Kerberos)

StephenSorriaux:
This adds the possibility to connect to Zookeeper using the GSSAPI mechanism a.k.a Kerberos.

__How to connect to ZK using GSSAPI?__
The following libs must be installed:
- [pure-sasl](https://github.com/thobbs/pure-sasl)
- [kerberos](https://github.com/apple/ccs-pykerberos)

This is an example that shows how connect to Zookeeper using Kerberos auth where test@EXAMPLE.COM is a Kerberos principal and kerberos.example.com the DNS to access to the kdc:
```
from kazoo.client import KazooClient
from kazoo.security import make_acl

kerberos_principal = 'test@EXAMPLE.COM'
zk = KazooClient(hosts='172.17.0.2:2181', auth_data=[('gssapi', kerberos_principal)], kerberos_server_principal='kerberos.example.com')
zk.start()
acl = make_acl('sasl', credential=kerberos_principal, all=True)
zk.ensure_path('/test', acl=(acl,))
zk.get('/test')
zk.stop()
```
`kerberos_principal` can be `None` if usage of Kerberos credential cache is wanted.

__How did you test this?__
Using a Docker container where Kerberos is configured and where Zookeeper is using it for authentication. Docker images are available in my repo https://github.com/StephenSorriaux/docker-kerberos-zookeeper
I didn't add any unit test for this new feature since it requires a running Kerberos server...

I updated the `add_auth` method documentation but it seems to me that this might not be the best place to put it since this method can not be used with `sasl` or `gssapi` schemes. I may move it elsewhere in another PR.

PR #546, opened 2018-12-11
feat(core): improve SASL interface

ceache:
feat(core): improve SASL interface

* Move SASL configuration out of auth_data into its own dictionary which
  exposes more SASL features (e.g. server service name, client principal,
  ...). Legacy syntax is still supported for backward compatibility.
* Remove SASL from auth_data and place it between 'connection' and
  'zookeeper protocol level authentication' to simplify connection logic
  and bring code in line with the protocol stack (SASL wraps Zookeeper,
  not the other way around).
* Consistent exception, `AuthFailedError`, raised during authentication
  failure between SASL and ZK authentication.
* New 'SASLException' exception raised in case of SASL intrisinc
  failures.
* Add support for GSSAPI (Kerberos).

Example connection using Digest-MD5:
```
  client = KazooClient(
      sasl_options={'mechanism': 'DIGEST-MD5',
                    'username': 'myusername',
                    'password': 'mypassword'}
  )
```

Example connection using GSSAPI (with some optional settings):

```
  client = KazooClient(
      sasl_options={'mechanism': 'GSSAPI',
                    'service': 'myzk',                  # optional
                    'principal': 'clt@EXAMPLE.COM'}     # optional
  )
```

I provided patches to pure-sasl to fix its DIGEST support but got pulled in by other projects before I could get around to clean it up, get legal approval, and contribute this back.
This is a rebase of code we have had in production for over a year now, with both DIGEST and GSSAPI authentications.

I have found this approach to be cleaner then trying to mix SASL data with Zookeeper's own auth_data. 

Disclaimer from my employer:

> THE FOLLOWING DISCLAIMER APPLIES TO ALL SOFTWARE CODE AND OTHER MATERIALS CONTRIBUTED IN CONNECTION WITH THIS SOFTWARE:
> 
> THIS SOFTWARE IS LICENSED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS “AS IS” AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE AND ANY WARRANTY OF NON-INFRINGEMENT, ARE DISCLAIMED.  IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.  THIS SOFTWARE MAY BE REDISTRIBUTED TO OTHERS ONLY BY EFFECTIVELY USING THIS OR ANOTHER EQUIVALENT DISCLAIMER IN ADDITION TO ANY OTHER REQUIRED LICENSE TERMS.
> 
> ONLY THE SOFTWARE CODE AND OTHER MATERIALS CONTRIBUTED IN CONNECTION WITH THIS SOFTWARE, IF ANY, THAT ARE ATTACHED TO (OR OTHERWISE ACCOMPANY) THIS SUBMISSION (AND ORDINARY COURSE CONTRIBUTIONS OF FUTURES PATCHES THERETO) ARE TO BE CONSIDERED A CONTRIBUTION.  NO OTHER SOFTWARE CODE OR MATERIALS ARE A CONTRIBUTION.
>


ceache:
As for testing, the same method outlined in #533 can be used.
I am also working on adding code to the testing harness to support setting up a test Kerb environment (without root)

anthonyliao:
This PR looks good. Was able to test this as well. See test below
@StephenSorriaux are you planning to merge this PR instead of #533 for GSSAPI authentication?

Test:

1. Create valid ticket in cache
```
$ kinit -l 5m anthonyliao
Password for anthonyliao:

$ klist
Ticket cache: FILE:/tmp/krb5cc_1524355
Default principal: anthonyliao

Valid starting       Expires              Service principal
12/7/2018 22:01:34  12/7/2018 22:06:32  tgt/testrealm.com@testrealm.com
```

2. Run the following script
```
from kazoo.security import make_acl
import time

zk = KazooClient(hosts='localhost', sasl_options={'mechanism':'GSSAPI'})
zk.start()

acl1 = make_acl('sasl', credential='anthonyliao', write=True)
acl2 = make_acl('world', credential='anyone', read=True, create=True, delete=True, admin=True)
zk.ensure_path('/TEST_NAMESPACE/testnode', acl=(acl1,acl2))
# Write some data to znode
zk.set('/TEST_NAMESPACE/testnode', 'data0')
# Update znode, will fail if not authenticated as anthonyliao
zk.set('/TEST_NAMESPACE/testnode', 'data1')
```

kazoo/protocol/connection.py
```diff
@@ -686,68 +674,112 @@ def _connect(self, host, port):
...
-    def _set_connected_ro_or_rw(self, client):
-        """ Called to decide whether to set the KeeperState to CONNECTED_RO
-            or CONNECTED"""
-        if self._ro:
-            client._session_callback(KeeperState.CONNECTED_RO)
-            self._ro_mode = iter(self._server_pinger())
-        else:
-            client._session_callback(KeeperState.CONNECTED)
-            self._ro_mode = None
+    def _authenticate_with_sasl(self, host, timeout):
```
StephenSorriaux: This was a possibility I thought of when developping the initial DIGEST-MD5 SASL connection but I am not sure if, while authenticating using SASL, the zk server can not send the client any other kind of messages... I also noticed that, in the official java zk client, the choice was made to handle SASL packets in the `readResponses()` loop (see https://github.com/apache/zookeeper/blob/master/zookeeper-server/src/main/java/org/apache/zookeeper/ClientCnxn.java#L876)

On your side, what do you think would be the best?

kazoo/protocol/connection.py
```diff
@@ -686,68 +674,112 @@ def _connect(self, host, port):
...
+        """Establish a SASL authenticated connection to the server.
+        """
+        if not PURESASL_AVAILABLE:
+            raise SASLException('Missing SASL support')
```
StephenSorriaux: As it is currently done, isn't it better to connect to the server without any auth and inform the client with a warning message?

kazoo/protocol/connection.py
```diff
@@ -643,7 +631,7 @@ def _connect(self, host, port):
 
         with self._socket_error_handling():
             self._socket = self.handler.create_connection(
-                address=(host, port),
+                address=(hostip, port),
```
StephenSorriaux: Did you notice issues when using the previous behaviour?

CHANGES.md
```diff
@@ -1,3 +1,14 @@
+<a name="2.7.0"></a>
+## 2.7.0 (next)
```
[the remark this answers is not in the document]
ceache: Ack will remove

kazoo/client.py
```diff
@@ -303,15 +307,6 @@ def _retry(*args, **kwargs):
         self.Semaphore = partial(Semaphore, self)
         self.ShallowParty = partial(ShallowParty, self)
 
-        # Managing SASL client
-        self.use_sasl = False
-        for scheme, auth in self.auth_data:
```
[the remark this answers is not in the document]
ceache: Ok, I will add depreation warning in client and populate the `sasl_options` config from the `auth_data`.

Thinking ahead, how do you want to handle the presence of both SASL data in `auth_data` and a `sasl_options` block? I am thinking I should error out.

kazoo/protocol/connection.py
ceache: My line of thought is that if I create a ZK client object with auth data and that auth data is rejected by the server, it feels to me that I provided "wrong data" and it should tell me right away.

Another way to look at this is that this is an "auth error" while trying to access a node you do not have permission to (which would likely happen if you failed to authenticate) is a "perm error".  Making the authentication failure silent would, in effect, alias the two kind of errors and make the ZK client less expressive.

Lastly, it would be trivial for an application to simply catch the auth error, reconfigure the ZK client *without* the bad auth data and reconnect, to emulate the silent continue.
On the other hand, it would not be as easy to connect and then check if we actually successfully authenticated.

That being said, in my work environment, authentication is pretty much mandatory, so my point of view might be skewed here.

kazoo/protocol/connection.py
ceache: This was early on. If I remember properly, without this, I was seeing IPs passed to the SASL library which was then creating principals in the form `zookeeper/1.2.3.4@myrealm` instead of `zookeeper/myhostname@myrealm`.

Do you have a concern with this?
As I said, this was a while ago, I can double check...

ceache:
@StephenSorriaux 

> Concerning connection using Kerberos, how are you managing the tickets renewal? It seems like the client must handle this, in case of re-connection

I do not try to implement a ticket cache that would try to leverage keytabs to refresh tickets. Java does this because they are going down the rabbit hole of trying to abstract the OS but this is far from trivial. You would have to consider all the options provided in the JAAS config (use cache, use keytabs etc), and have some tooling to understand the expiration and renewals options of tickets.

Full disclosure: we have a completely managed environment in this regard so we did not need this functionality at work.
If we wanted to add this though, shouldn't it be implemented in a separate library than kazoo?

kazoo/client.py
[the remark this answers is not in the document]
StephenSorriaux: I agree, erroring out is great.

kazoo/protocol/connection.py
StephenSorriaux: You are right, it makes sense, let's keep it the way you did.

kazoo/protocol/connection.py
StephenSorriaux: Not much concern, just wanting to understand the reason for the change and keep some history about it.

StephenSorriaux:
> @StephenSorriaux
> 
> > Concerning connection using Kerberos, how are you managing the tickets renewal? It seems like the client must handle this, in case of re-connection
> 
> I do not try to implement a ticket cache that would try to leverage keytabs to refresh tickets. Java does this because they are going down the rabbit hole of trying to abstract the OS but this is far from trivial. You would have to consider all the options provided in the JAAS config (use cache, use keytabs etc), and have some tooling to understand the expiration and renewals options of tickets.
> 
> Full disclosure: we have a completely managed environment in this regard so we did not need this functionality at work.
> If we wanted to add this though, shouldn't it be implemented in a separate library than kazoo?

I agree with you, you certainly most know about Kerberos than me. Maybe we could also those informations in the documentation in order to help any user for the settings?

kazoo/client.py
```diff
@@ -123,6 +123,9 @@ def __init__(self, hosts='127.0.0.1:2181',
             A list of authentication credentials to use for the
             connection. Should be a list of (scheme, credential)
             tuples as :meth:`add_auth` takes.
+        :param sasl_options:
```
StephenSorriaux: In order to help the user during setting, can you add the possible keys for this dict?

kazoo/protocol/connection.py
ceache: I have studied the Java client code and it seems to me that no other packet is allowed while SASL negotiation is in progress. Their code goes to great length to bypass all send queue during that process and that the packets are strict XID for XID request/response exchanges: 
- https://github.com/apache/zookeeper/blob/caca062767c36525e6ecead2ae0f34c447394809/zookeeper-server/src/main/java/org/apache/zookeeper/ClientCnxnSocketNIO.java#L170
- https://github.com/apache/zookeeper/blob/master/zookeeper-server/src/main/java/org/apache/zookeeper/ClientCnxn.java#L890
- https://github.com/apache/zookeeper/blob/cb9f303bda9137d1aebe8eff3eab85c8a59f3cdd/zookeeper-server/src/main/java/org/apache/zookeeper/client/ZooKeeperSaslClient.java#L296
 

Furthermore, the client also does check that the state is `CONNECTED` or `CONNECTEDREADONLY` *before* starting the SASL authentication:

- https://github.com/apache/zookeeper/blob/master/zookeeper-server/src/main/java/org/apache/zookeeper/ClientCnxn.java#L1163
- https://github.com/apache/zookeeper/blob/22e4acc802904231d537ae18f479cfc729851d87/zookeeper-server/src/main/java/org/apache/zookeeper/ZooKeeper.java#L638

This indicate the the SASL negotiation only starts once the connection to the server is established (which means a `CONNECTED` event was already emitted), which my PR emulates. Note that native forms of authentication, when passed in as `auth_data`, are also evaluated right after the client reaches the connected state.
The only possible downside to my approach is a connection watcher would get a `CONNECTED` event followed by a `LOST` which is already possible (networks can fail) and what happens with native authentication schemes.
Looking at the links above, what do you think? Do you agree with my interpretation?

As for factoring out the SASL negotiation in its own function, since only SASL packets can be sent during the authentication process anyway, and only during connection, I do not see how this could bite us down the road  (and it definitely made my PR less invasive).
If you still feel strongly about this, can you give me guidance as to how you would tackle integrating the SASL loop, with the above state conditions, in the main send/receive loop?
Also, could this be done in a different refactoring PR?

kazoo/client.py
ceache: done

kazoo/protocol/connection.py
StephenSorriaux: Thanks for all those links, you are right, I missed the `findSendablePacket` method and it really seems nothing can happen during the SASL authentication. It makes totally sense to do the things the way you suggest.

StephenSorriaux:
Thanks a lot for this PR and for your awesome work.

StephenSorriaux:
Merged to master, and might be a 2.7.0 Kazoo version on its own.

anthonyliao:
Hi @StephenSorriaux, are there any plans to cut a release for this feature soon?

StephenSorriaux:
Hi @anthonyliao, I would like to also merge https://github.com/python-zk/kazoo/pull/547 for the 2.7.0 version. Unfortunately, I did not find time to review / update it for now...

ceache:
@StephenSorriaux @anthonyliao I would like to investigate the high rate of "timeouts" in the Travis integration jobs too.
This is not particularly related to SASL functions, or #547 but something we might want to check before cutting a new release.

I *think* I traced it to the `ZkClient.stat()` code assuming any early error is a "timeout"  but I am still investigating.
I have been quite busy recently at work but I'll try to clear this this weekend.

PR #547, opened 2018-12-18
 feat(core): run SASL Kerberos tests as part of build

ceache:
* Install debian packages for KDC as part of Travis init.
 * Setup a loopback mini KDC for running tests.
 * Run SASL tests as part of Travis builds.
 * Improve harness cluster to support:
    * Reconfiguration when environment changes.
    * Different JAAS configurations (DIGEST/GSSAPI).
 * Moved SASL tests into own module, with specially configured harness.
 * Bumped default timeout to 15 sec to mitigate false negatives on
   Travis.


StephenSorriaux:
Thanks for your PRs. I will try to review it by the end of the week.

anthonyliao:
Hi @ceache, any plans to look into this? I believe @StephenSorriaux would like to merge this PR before releasing 2.7.0. Thanks

StephenSorriaux:
I took a lot of times to test this, and we still have issues when trying to raise the `AuthFailedError` at `client.start()`. Sometimes it works because we have some thread that regularly sends a ping request to the server that triggers the the `AuthFailedError`, but sometimes it ends up with a `KazooTimeoutError` raised by the `client.start()` method. I think we need to tackle this before releasing a new version.

ceache:
@stephensorriaux would you be interested in going the docker route for the KDC?
I have expanded on the `init_krb5` script I provided here into a reusable docker image. https://github.com/ceache/Dockerfiles/blob/master/README.md 

It would avoid repeating the issues that stemmed from the Xenial upgrade

StephenSorriaux:
Hi @ceache,

I think the Docker route will be a great addition to the way we currently test our lib but I am not sure it would help to fix the current issue. The Xenial upgrade "silently" changed the java version used from 8 to 11 and makes it impossible to connect the ZK server to the KDC due to some `Message stream modified (41)` error. I tried several things (changing ciphers, updating the configuration, etc.) and managed to find that the issue appears due to some changes between Java 8 and Java 9, but no real solutions for now since it is a painful error to debug. I think I will just force the Java version to 8 via the Travis configuration so that I can merge this and release a new version of Kazoo.
If you have already encountered a similar issue or have any ideas, I would very appreciate a little help.

ceache:
I have never encountered this error before but I'll have a look.

kazoo/tests/test_sasl.py
```diff
@@ -0,0 +1,186 @@
...
+from kazoo.testing import KazooTestHarness
+from kazoo.exceptions import (
+    AuthFailedError,
+    NoAuthError,
+)
+from kazoo.tests.util import TRAVIS_ZK_VERSION
+
+
+class TestLegacySASLDigestAuthentication(KazooTestHarness):
+
```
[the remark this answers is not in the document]
ceache: Addressed

ceache:
@StephenSorriaux I think I got it. At least now it passes the SASL/GSSAPI tests.
Still see some random timeout errors though...

StephenSorriaux:
Thanks @ceache. Would you have some time to update your branch? Then lets put all of this on master.

ceache:
@StephenSorriaux rebased on top of master

init_krb5.sh
```diff
@@ -78,6 +78,8 @@ cat <<EOF >${WRK_DIR}/krb5.conf
   kdc_listen = 127.0.0.1:${KDC_PORT}
   kdc_tcp_listen = 127.0.0.1:${KDC_PORT}
   kdc = 127.0.0.1:${KDC_PORT}
+  kdc_ports = ${KDC_PORT}
+  kdc_tcp_ports = ""
   default_domain = KAZOOTEST.ORG
  }
```
jeffwidman: FYI typo in commit message here: "desactivate" --> "deactivate"

Doesn't matter if this will all get squashed, but looking at your clean commit history of distinct changes I'd rather not squash these as it's helpful to keep them as separate IMO

jeffwidman:
FYI typo in commit message on 3f32901d75d031eeebc4882562c0b249b590cc68: "desactivate" --> "deactivate"

Doesn't matter if this will all get squashed, but looking at your clean commit history of distinct changes with clear commit messages I'd rather not squash these as it's helpful to keep them as separate IMO

kazoo/testing/harness.py
```diff
@@ -16,52 +16,105 @@
...
-            install_path=ZK_HOME,
-            classpath=ZK_CLASSPATH,
-            port_offset=ZK_PORT_OFFSET,
-            size=ZK_CLUSTER_SIZE,
-            observer_start_id=ZK_OBSERVER_START_ID,
-            configuration_entries=additional_configuration_entries,
-            java_system_properties=additional_java_system_properties
-        )
-        atexit.register(lambda cluster: cluster.terminate(), CLUSTER)
+            log.info('Config change detected. Reconfigurating cluster...')
```
jeffwidman: typo? Reconfigurating --> Reconfiguring

test-requirements.txt
```diff
@@ -0,0 +1,5 @@
+coverage==3.7.1
+mock==1.0.1
+nose==1.3.3
+flake8==2.3.0
+objgraph==3.4.0
```
jeffwidman: 👍 thanks for cleaning this up.

However, I personally prefer to call it `requirements-test.txt` so that if viewing a folder, all the `requirements*.txt` files will be next to each other. Otherwise very easy to overlook that there are separate files.

ceache: Done

kazoo/testing/harness.py
ceache: fixed

jeffwidman:
Tests are now passing @ceache want to merge it? Be nice to land this (assuming it's ready)

PR #549, opened 2019-01-07
fix(handlers): make AsyncResult call all registered callbacks instantly if the handler has stopped running

laura-surcel:
Hi!

I'm using kazoo in a work-related project and sometimes our tests are leaving zombie threads that are blocked in `self._condition.wait(timeout)` from `AsyncResult:get`. 

After some digging it looks like the problem occurs when we start a KazooClient, try to make an operation on it like a path creation and then stop it fast. In this case, the callbacks of the `AsyncResult` may not be called if `AsyncResult:set`/`AsyncResult:set_exception` are called just after the client is stopped. As I said it might not happen too often for most people but for us it's like in 10% of the test runs and it's slows  down our development pipeline. I've made a custom `AyncResult` class in our project that modifies a bit the behavior and seems to fix the issue for us but maybe you guys would also consider this solution for the official version.

If you have other suggestions or concerns let me know.

### DIFF
--- a/kazoo/handlers/utils.py
+++ b/kazoo/handlers/utils.py
@@ -114,17 +114,17 @@ class AsyncResult(object):
                 self._callbacks.remove(callback)
 
     def _do_callbacks(self):
-        leftovers = set()
+        """Execute the callbacks that were registered by :meth:`rawlink`.
+        If the handler is in running state this method only schedules
+        the calls to be performed by the handler. If it's stopped,
+        the callbacks are called right away."""
 
         for callback in self._callbacks:
             if self._handler.running:
                 self._handler.completion_queue.put(
                     functools.partial(callback, self))
             else:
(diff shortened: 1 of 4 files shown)


kazoo/handlers/utils.py
```diff
@@ -122,6 +113,18 @@ def unlink(self, callback):
             if callback in self._callbacks:
                 self._callbacks.remove(callback)
 
+    def _do_callbacks(self):
```
StephenSorriaux: Can you please add a little description to this method? Just to help make the read straightforward for any newcomer.

laura-surcel: Sure. I'll add one.

laura-surcel: Done.

StephenSorriaux:
Merged to master.

PR #559, opened 2019-05-02
feat(recipe): add support for curator SharedCount recipe

### LINKED ISSUE #558: Allow counter recipe to work seamlessly with curator's SharedCount recipe
Today the kazoo counter recipe stores the value using ascii encoding while the curator SharedCount recipe stores the value using a signed 4 byte representation in big Endian byte order. Therefore, if a java client using curator and a python client using kazoo need to work with the same counter they will receive type errors.

I would like to submit a pull request that allows the two clients to co-exist peacefully. 

The idea is to add a flag to kazoo's counter recipe constructor that instructs it to read and store data the same way that curator does. This option will default to off so as not to affect any existing clients that do not need this functionality.

Kazoo supports both int and float; however, curator's SharedCount only supports int. Therefore, whenever support for curator is enabled kazoo will only allow int values.

Since python 2.7 is nearing its end of life, this change must be compatible with both python 2 and 3.

Example use:
`counter = zk.Counter("/curator", support_curator=True) `
`counter += 2 `
`counter -= 1 `
`counter.value == 1 `
`counter.pre_value == 2 `
`counter.post_value == 1`

BrianEaton1:
this patch allows java clients using curator's SharedCount recipe
and python clients using kazoo's Counter recipe to read and
write from the same path without receiving type errors.

example use:
`counter = zk.Counter("/curator", support_curator=True)`
`counter += 2`
`counter -= 1`
`counter.value == 1`
`counter.pre_value == 2`
`counter.post_value == 1`

Closes #558

### DIFF
--- a/kazoo/recipe/counter.py
+++ b/kazoo/recipe/counter.py
@@ -62,7 +62,7 @@ class Counter(object):
         :param client: A :class:`~kazoo.client.KazooClient` instance.
         :param path: The counter path to use.
         :param default: The default value to use for new counter paths.
-        :param support_curator: Enable if support for curator's SharedCount 
+        :param support_curator: Enable if support for curator's SharedCount
                                 recipe is desired.
 
         """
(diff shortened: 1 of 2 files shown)


StephenSorriaux:
Hi @BrianEaton1,

Thanks a lot for your time to produce this PR. I will review it as soon as possible.

PR #568, opened 2019-07-31
Extend create_tcp_connection utility

Creatone:
It would be a great thing to set context options and ciphers to provide better security.


StephenSorriaux:
Hi,

Thanks for your PR. I agree this would be a good feature. However, I noticed you only changed one method, so, for instance, the client interface will not have a way to set those parameters. Am I missing something here?

Creatone:
Client can set this parameters in his handler.
For example:
```
class MySequentialThreadingHandler(SequentialThreadingHandler):
    def create_connection(self, *args, **kwargs):
        return create_tcp_connection(socket, options=MY_OPTIONS, ciphers=MY_CIPHERS,
                                     *args, **kwargs)
```

StephenSorriaux:
On master

Issue #374, opened 2015-12-28
Incorrect SessionExpiredError when ZooKeeper leader process down


bwtakacy:
I encountered SessionExpiredError when ZooKeeper leader process down with below stack trace, but there was no session expiration at that time in ZooKeeper server logs:

```
Exception in thread Thread-6:
Traceback (most recent call last):
  File "/usr/lib64/python2.6/threading.py", line 532, in __bootstrap_inner
    self.run()
  File "/home/postgres/zha/zha.py", line 284, in run
    self.in_elector_loop()
  File "/home/postgres/zha/zha.py", line 303, in in_elector_loop
    lock_result = self.lock.acquire(timeout=self.zha.config.get("elector_interval",3))
  File "/usr/lib/python2.6/site-packages/kazoo-2.2.1-py2.6.egg/kazoo/recipe/lock.py", line 174, in acquire
    self._delete_node(self.node)
  File "/usr/lib/python2.6/site-packages/kazoo-2.2.1-py2.6.egg/kazoo/recipe/lock.py", line 266, in _delete_node
    self.client.delete(self.path + "/" + node)
  File "/usr/lib/python2.6/site-packages/kazoo-2.2.1-py2.6.egg/kazoo/client.py", line 1295, in delete
    return self.delete_async(path, version).get()
  File "/usr/lib/python2.6/site-packages/kazoo-2.2.1-py2.6.egg/kazoo/handlers/utils.py", line 72, in get
    raise self._exception
SessionExpiredError
```

, where zha.py is a tools using kazoo.

After debugging, I have found that _call() of KazooClient raised this error when KeeperState became CONNECTING and this is what happened in my environment.

```
        elif self._state in (KeeperState.EXPIRED_SESSION,
                             KeeperState.CONNECTING):
            async_object.set_exception(SessionExpiredError())
            return False
```

Is there some historical reason about this?
Is it OK to change the condition will catch only the KeeperState.EXPIRED_SESSION ?

harlowja:
Said change seems ok although I'm not sure on the history of why it checks against both.

bwtakacy:
Thank you for take a look!
I will create a patch and test it.

bwtakacy:
I tried to remove checking KeeperState.CONNECTIONG, but tests failed because the below test in test_client.py assumes that.

```
 468     def test_create_on_broken_connection(self):
 469         client = self.client
 470         client.start()
 471 
 472         client._state = KeeperState.EXPIRED_SESSION
[...]

PR #570, opened 2019-08-28
Fix bug where kazoo client returns SESSION EXPIRED when in CONNECTING state

prajay:
kazoo client returns a SessionExpired when its state is in CONNECTING. This is not an indication of session expiry and could just be that the client is still re-establishing a connection to the server after a network blip. Hence, the error returned should be a ConnectionLoss and not a SessionExpired

### DIFF
--- a/kazoo/client.py
+++ b/kazoo/client.py
@@ -595,9 +595,11 @@ class KazooClient(object):
             async_object.set_exception(ConnectionClosedError(
                 "Connection has been closed"))
             return False
-        elif self._state in (KeeperState.EXPIRED_SESSION,
-                             KeeperState.CONNECTING):
+        elif self._state == KeeperState.EXPIRED_SESSION:
             async_object.set_exception(SessionExpiredError())
+            return False
+        elif self._state == KeeperState.CONNECTING:
+            async_object.set_exception(ConnectionLoss())
             return False
 
         self._queue.append((request, async_object))
--- a/kazoo/tests/test_client.py
+++ b/kazoo/tests/test_client.py
@@ -593,7 +593,7 @@ class TestClient(KazooTestCase):
                           '/closedpath', b'bar')
 
         client._state = KeeperState.CONNECTING
-        self.assertRaises(SessionExpiredError, client.create,
+        self.assertRaises(ConnectionLoss, client.create,
                           '/closedpath', b'bar')
         client.stop()
         client.close()


StephenSorriaux:
Hi,

Thanks for this PR, I will link into it. Don't bother with the tests for now, I will re-trigger the Travis build.

prajay:
Hey @StephenSorriaux,

Just wanted to check on this. Does the change seem reasonable to you?

StephenSorriaux:
Hey @prajay,

I did not have time to go deep in the reason why we chose to raise `SessionExpiredError` in this case. Also, this change is not that straightforward since some users might be catching this exception for their use case, so ,as if, changing it to `ConnectionLoss` would be a breaking change.

ztzg:
Hi @StephenSorriaux, @prajay,

A note to let you know that we are also seeing the issue addressed by this PR (also reported as
https://github.com/python-zk/kazoo/issues/374) and would like to see something akin to this patch in Kazoo `master`.

The main issue is that spurious "session expired" exceptions can cause applications to tear down much more state than required, and to try to rebuild it—further stressing an ensemble which may just have, for example, lost its leader.

I see that the goal of the commit (by @bbangert) which introduced the `CONNECTING` check was to allow applications to distinguish between a method invoked on a `CLOSED` client and a server-generated `SESSIONEXPIRED` (error code -112): https://github.com/python-zk/kazoo/commit/4eabbd1d8fffe31d25904bf456687c48c3b19848

I basically see three ways of allowing applications to discriminate between these different states while retaining "full" backward compatibility:

 1. No code change, but a "redefinition" of "session expired" to `SessionExpiredError` + `client.client_state == KeeperState.EXPIRED_SESSION`.

 2. Keeping the `KeeperState`, or alternatively, "fake codes" in the existing exception. Something akin to:

    ```python
    @_zookeeper_exception(-112)
    class SessionExpiredError(ZookeeperError):
        def __init__(self, *, state=None):
            self.state = state
    ```

 3. Taking inspiration of the commit above: introducing distinct subclasses of `SessionExpiredError` (one per actual condition) and redirecting `EXCEPTIONS[-112]` to one of them.

Each of these options is "distasteful" in its own way, but I could nevertheless whip up patches/PRs for #3 and/or #2 if you think they would be worth a closer look.

(#3 would probably be more "natural" if we manage to find "reasonable" subclass names.  `ConnectingError` and `ServerSessionExpiredError` are the best I could come up with so far.  Suggestions are welcome!)

StephenSorriaux:
Hi @ztzg,

Thanks for taking some time to share your thoughts about this subject.

I understand what you want to do the and I start to think that the `CONNECTING` state should actually not raise any exception but, instead, just makes the request be queued as it is done with the `CONNECTED` state. Since the request stays in the queue as long as no connection is available, it should work fine but that definitely needs some testing (no side-effect on recipes, ensure that we don't keep too much requests, etc.).

ztzg:
Not raising an exception on `CONNECTING` does indeed sound like the best solution (and matches what other clients are doing).

I had written it off as requiring invasive changes and/or going against the "philosophy" of Kazoo (no `setWatches`, etc.).  I could have a closer look and see if I can come up with a proof of concept which is not too disruptive, if you think it has a chance to move forward.

StephenSorriaux:
I really would like to more forward with this issue, so yes any contributions is very welcomed.

jeffwidman:
Nudge @ztzg

ztzg:
@jeffwidman: Still in my TODO list, but other things got on top.  You're right that I should move forward a bit on this front, though.  I'll keep your nudge in mind! :)

StephenSorriaux:
Thanks @prajay for your time on this PR, I am closing it in favor of https://github.com/python-zk/kazoo/pull/588.

PR #583, opened 2020-02-09
WIP fix(core): allow requests to be queued in CONNECTING state (#374)

ztzg:
As discussed in https://github.com/python-zk/kazoo/pull/570#issuecomment-554798550:

With this patch, requests issued while the client is in the `CONNECTING` state get queued instead of raising a (misleading) `SessionExpiredError`.

TBD:

* [X] This patch does not prevent requests which have been queued but not emitted from being rejected with `ConnectionLoss`.  It should also get rid of the second part of `_notify_pending`, shouldn't it? [Not doing this, as the C/Java clients do not, either.] 

* [x] This patch does not try to limit the maximum queue length, which should probably be controlled via a new `KazooClient` parameter.  (How about `max_queue_length`?  What should it default to?  Which exception should we raise when the queue overflows?) [Not doing this, as the C/Java clients do not, either.]

### DIFF
--- a/kazoo/client.py
+++ b/kazoo/client.py
@@ -584,7 +584,7 @@ class KazooClient(object):
         the queue if there is.
 
         Returns False if the call short circuits due to AUTH_FAILED,
-        CLOSED, EXPIRED_SESSION or CONNECTING state.
+        CLOSED, or EXPIRED_SESSION state.
 
         """
 
@@ -595,8 +595,7 @@ class KazooClient(object):
             async_object.set_exception(ConnectionClosedError(
                 "Connection has been closed"))
             return False
-        elif self._state in (KeeperState.EXPIRED_SESSION,
-                             KeeperState.CONNECTING):
+        elif self._state == KeeperState.EXPIRED_SESSION:
             async_object.set_exception(SessionExpiredError())
             return False
 
--- a/kazoo/client.py
+++ b/kazoo/client.py
@@ -561,13 +561,14 @@ class KazooClient(object):
             except IndexError:
                 break
 
-        while True:
-            try:
-                request, async_object = self._queue.popleft()
-                if async_object:
-                    async_object.set_exception(exc)
-            except IndexError:
-                break
+        if state != KeeperState.CONNECTING:
+            while True:
+                try:
+                    request, async_object = self._queue.popleft()
+                    if async_object:
+                        async_object.set_exception(exc)
+                except IndexError:
+                    break
 
     def _safe_close(self):
         self.handler.stop()
--- a/kazoo/protocol/connection.py
+++ b/kazoo/protocol/connection.py
@@ -59,8 +59,8 @@ log = logging.getLogger(__name__)
 
 # Special testing hook objects used to force a session expired error as
 # if it came from the server
-_SESSION_EXPIRED = object()
-_CONNECTION_DROP = object()
+_SESSION_EXPIRED = (SessionExpiredError, "Session expired: Testing")
+_CONNECTION_DROP = (ConnectionDropped, "Connection dropped: Testing")
 
 STOP_CONNECTING = object()
 
@@ -454,12 +454,13 @@ class ConnectionHandler(object):
                 pass
             return
 
-        # Special case for testing, if this is a _SessionExpire object
-        # then throw a SessionExpiration error as if we were dropped
-        if request is _SESSION_EXPIRED:
-            raise SessionExpiredError("Session expired: Testing")
-        if request is _CONNECTION_DROP:
-            raise ConnectionDropped("Connection dropped: Testing")
+        # Special case for testing, if this is a _SESSION_EXPIRED or
+        # _CONNECTION_DROP object, throw the corresponding error.
+        if request is _SESSION_EXPIRED or request is _CONNECTION_DROP:
+            client._queue.popleft()
+            self._read_sock.recv(1)
+            error, arg = request
+            raise error(arg)
 
         # Special case for auth packets
         if request.type == Auth.type:
--- a/kazoo/tests/test_client.py
+++ b/kazoo/tests/test_client.py
@@ -592,14 +592,66 @@ class TestClient(KazooTestCase):
         self.assertRaises(AuthFailedError, client.create,
                           '/closedpath', b'bar')
 
-        client._state = KeeperState.CONNECTING
-        self.assertRaises(SessionExpiredError, client.create,
-                          '/closedpath', b'bar')
         client.stop()
         client.close()
 
         self.assertRaises(ConnectionClosedError, client.create,
                           '/closedpath', b'bar')
+
+    def test_create_queued_while_connecting(self):
+        # This is a bit tricky: we are trying to show that a request
+        # can be queued while the connection is down.  So we create a
+        # client with a fairly large but stable reconnect delay, and
+        # perform a number of attempts.
+        handler = self._makeOne()
+        sleep_func = handler.sleep_func
+        client = self._get_client(
+            handler=handler,
+            connection_retry=dict(
+                max_tries=-1,
+                delay=0.5,
+                backoff=1,
+                max_jitter=0.0,
+                sleep_func=sleep_func
+            )
+        )
+        client.start()
+
+        max_attempts = 5
+        attempts = 0
+        was_queued = False
+
+        while not was_queued and attempts < max_attempts:
+            attempts += 1
+
+            try:
+                client.delete('/queued')
+            except NoNodeError:
+                pass
+
+            # Shut the socket down, and wait for the client to
+            # transition to non-connected.
+            client._connection._socket.shutdown(socket.SHUT_RDWR)
+            while client.connected:
+                sleep_func(0.001)
+
+            # Issue an async request, assuming the client hasn't had
+            # enough time to reconnect.
+            result = client.create_async('/queued', None)
+
+            # Consider the test potentially satisfactory if client
+            # still hasn't had time to reconnect.
+            was_queued = not client.connected
+
+            # Wait for expected result.
+            self.assertEqual(result.get(), '/queued')
+            self.assertEqual(client.connected, True)
+
+        # Fail if no "window" was observed despite max_attempts.
+        self.assertEqual(was_queued, True)
+
+        client.stop()
+        client.close()
 
     def test_create_null_data(self):
         client = self.client
--- a/kazoo/tests/test_connection.py
+++ b/kazoo/tests/test_connection.py
@@ -7,11 +7,10 @@ import struct
 import sys
 
 from nose import SkipTest
-from nose.tools import eq_
-from nose.tools import raises
+from nose.tools import eq_, raises, assert_not_equal
 import mock
 
-from kazoo.exceptions import ConnectionLoss
+from kazoo.exceptions import ConnectionLoss, SessionExpiredError
 from kazoo.protocol.serialization import (
     Connect,
     int_struct,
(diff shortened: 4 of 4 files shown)


ztzg:
@jeffwidman, @StephenSorriaux: As far as I can tell, just removing the check for `CONNECTING` from `KazooClient._call` is sufficient to obtain the desired behavior; `_queue` and `_write_sock` should remain valid in that state (with the exception of the "race with close()" which is already handled by the code). Do you agree?

jeffwidman:
First, I have not looked deeply at this section of the code, so please take what I say with a grain of salt.

> removing the check for CONNECTING from KazooClient._call is sufficient to obtain the desired behavior; _queue and _write_sock should remain valid in that state (with the exception of the "race with close()" which is already handled by the code). Do you agree?

That sounds very reasonable. 

> This patch does not prevent requests which have been queued but not emitted from being rejected with ConnectionLoss. It should also get rid of the second part of _notify_pending, shouldn't it?

I'm not sure. @StephenSorriaux can you also take a look at this? It's a very small code change, but connection blips are common enough that we do want to make sure we get the model right.

> This patch does not try to limit the maximum queue length, which should probably be controlled via a new KazooClient parameter. (How about max_queue_length? What should it default to? Which exception should we raise when the queue overflows?)

How do the Java and C clients handle this? IMO they're the canonical examples so most people will be expecting similar behavior...

ztzg:
> > This patch does not try to limit the maximum queue length, which should probably be controlled via a new KazooClient parameter. (How about max_queue_length? What should it default to? Which exception should we raise when the queue overflows?)
> 
> How do the Java and C clients handle this? IMO they're the canonical examples so most people will be expecting similar behavior...

From the top of my head: they just queue requests, and don't have a maximum queue length.  (I will double-check when I circle back to this.)

ztzg:
> > This patch does not try to limit the maximum queue length, which should probably be controlled via a new KazooClient parameter. (How about max_queue_length? What should it default to? Which exception should we raise when the queue overflows?)
> 
> How do the Java and C clients handle this? IMO they're the canonical examples so most people will be expecting similar behavior...

  * Java: The client simply queues the request "packet" on a `LinkedBlockingQueue` of capacity `Integer.MAX_VALUE` (2^31 - 1):

    https://github.com/apache/zookeeper/blob/517ecde9dd3cfe574118231ec242ea73b80a4e79/zookeeper-server/src/main/java/org/apache/zookeeper/ClientCnxn.java#L533

    It thus (indirectly) throws an `IllegalStateException("Queue full")` after queuing 2+ billions of packets.

  * C: Same, except that the client uses an unbounded intrusive linked list, returning `ZSYSTEMERROR` if it fails to allocate memory:

    https://github.com/apache/zookeeper/blob/ae1a52c75b19e15db5913ac34d34b1c6fbf75e36/zookeeper-client/zookeeper-client-c/src/zookeeper.c#L1645-L1652

Neither checks the state of the connection before doing so.

jeffwidman:
Sounds good, we can do the same. That handles one of your outstanding TODOs.

For the second part:
> This patch does not prevent requests which have been queued but not emitted from being rejected with ConnectionLoss. It should also get rid of the second part of _notify_pending, shouldn't it?

Can you explain this a little more?
On the surface it sounds like a good thing if requests that were queued but then not emitted due to a permanent network failure eventually raise a ConnectionLoss... but we'd want to make sure some timeout had passed to guarantee permanent failure. Because if it was a temp network blip that the library recovers successfully and then sends these messages, then obviously we want to mark as successful. 

Sorry I'm asking for all these details, I'm trying to be somewhat thoughtful on a limited time budget and I know you've looked at this code deeply so already understand it.

Looking forward to landing this soon!

ztzg:
> Sounds good, we can do the same. That handles one of your outstanding TODOs.

Seems reasonable.

> For the second part:
> 
> > This patch does not prevent requests which have been queued but not emitted from being rejected with ConnectionLoss. It should also get rid of the second part of _notify_pending, shouldn't it?
> 
> Can you explain this a little more?

Sure.

> On the surface it sounds like a good thing if requests that were queued but then not emitted due to a permanent network failure eventually raise a ConnectionLoss... but we'd want to make sure some timeout had passed to guarantee permanent failure. Because if it was a temp network blip that the library recovers successfully and then sends these messages, then obviously we want to mark as successful.

First, something I am not planning to change: when the connection is lost, the requests which have been emitted but are not completed fail with `ConnectionLoss` (as the ACKs will never reach the client anyway).

With the current version of the patch, the requests which were queued but *not* emitted are also cancelled with the same exception:

https://github.com/python-zk/kazoo/blob/3a1d442ee77254bb6f03c70912b1520ed71f6906/kazoo/client.py#L564-L570

The idea would be not to cancel these requests when the new state is `CONNECTING`—as the client is trying to validate an existing session, and these "pristine" requests could very well be submitted in-order over the new connection if the session is recovered.

(Moreover, this would be more consistent with the rest of the patch; it would not matter if a request was submitted just before or during the "outage"—as long as it was not emitted.)

I have implemented the feature, but it currently breaks some assumptions in the tests; I will append a a second commit to the PR once I get to fix that.  (And it is fine with me if we decide to only merge the first half in the end.)

> Sorry I'm asking for all these details, I'm trying to be somewhat thoughtful on a limited time budget and I know you've looked at this code deeply so already understand it.

No problem! I have looked somewhat deeply, but am not very familiar with the codebase—so I am happy that you are questioning my assumptions.

jeffwidman:
Thanks, that is helpful.

I definitely agree with you that queued but not emitted should not fail while in `CONNECTING` state... essentially we want that reconnection to be invisible to the user (other than error logs). 

The one thing I do want to make sure of is that if the connection never manages to recover, all the queued-but-not-emitted requests are in some way are visible as either failed or incomplete to the application. I think you understand why, but if not let me know and I can explain further.

Assuming that is true, then I am in agreement with everything you said.

Thank you again!

ztzg:
> The one thing I do want to make sure of is that if the connection never manages to recover, all the queued-but-not-emitted requests are in some way are visible as either failed or incomplete to the application.

ACK; I think we are on the same page :)

StephenSorriaux:
Thank you for this PR and sorry for this late reply.

I agree with both of you on the fact that the `_queue` should not be emptied if the connection was somehow temporary lost.

Do you think it would be possible to precise this "new" behavior somewhere in the documentation?

ztzg:
@StephenSorriaux wrote:

> Thank you for this PR and sorry for this late reply.

No problem!

> I agree with both of you on the fact that the `_queue` should not be emptied if the connection was somehow temporary lost.

@jeffwidman, @StephenSorriaux: I believe the first commit of this series (the actual fix for https://github.com/python-zk/kazoo/issues/374), should be good to go.  It does not try to bound the queue length, but neither do the Java or C clients.

I have now implemented the new feature (not emptying the queue on retry-able disconnect), and have fixed/augmented the tests—but am now inclined to think that it should **not** be merged!  (Hence the `WIP` in the commit title.)

Indeed, it turns out that both the Java and C clients cancel such requests when the connection is lost:
    
  * Java: `ClientCnxn` socket error → `cleanAndNotifyState` → `cleanup` → `conLossPacket`/`remove`;
    
  * C: Socket error → `handle_socket_error_msg` → `handle_error` → `cleanup` → `cleanup_bufs` → `free_buffers`/`free_completions`.

I have double-checked; this happens even when the errors are not fatal and the session is recovered. (The C client doesn't even have a way to distinguish between queued & pending, and thus naturally cancels everything.)

I'm not sure it makes a difference in practice, as there is a "race" between the queuing of requests and the session event notification—particularly in the Java version which does not even lock the `outgoingQueue`—and one cannot assume that every request submitted before the notification is going to be canceled.  But staying aligned with the Java/C clients might just be a better idea.

Anyway: I have pushed both patches, and will let you have a look.  Unless you disagree, I will respin without the second one in a few days.

> Do you think it would be possible to precise this "new" behavior somewhere in the documentation?

Sure; I'll add a note for each patch we end up including.

jeffwidman:
Oh boy, that's a doozy of a research report. Thank you for digging into this. If you ever get tired of consulting and want a full time gig, I know a number of folks who are always looking to hire engineers with such attention to detail. 😁 

So I think you're on the right track here... on the surface at least, it seems the Java and C clients are making a mistake. But I also agree we should not deviate from them because most knowledgable ZK folks tend to be experts in ZK who switch-hit in polyglot environments so it's easier for the ecosystem if the behavior stays consistent across clients, even if it's arguably slightly incorrect (but clearly workable otherwise it would have been fixed long ago). 

So I think we should do as you suggest:
1. open a second PR with just the first fix
2. keep the second commit in this PR around (ie, don't force-push and obliterate it), but close it for now so it doesn't get merged.

Furthermore, I think we should go one step further and submit an issue to the upstream ZK team to change this behavior in the Java and C clients.

Do you want to submit this or shall I? If you don't have time I'm willing to do it if, but you discovered the issue, you understand it far better than I do, and as a consultant it's always a nice trust builder with potential clients to say that you've got a patch accepted to a core project like Zookeeper. Up to you, I just want to make sure in some way this is pushed upstream which which should result in either a fix or further clarification.

As a heads up, my personal experience with the core ZK project is that it tends to be a slower moving project.

Thoughts?

ceache:
@jeff Damien knows, he already has helped get SASL support in the C client
😉

ztzg:
> Oh boy, that's a doozy of a research report. Thank you for digging into this. If you ever get tired of consulting and want a full time gig, I know a number of folks who are always looking to hire engineers with such attention to detail. 

:)  I'm just trying to be a bit careful with concurrent systems programming; I know that one never is *too* prudent with those.  But thank you for the kind words !

> So I think you're on the right track here... on the surface at least, it seems the Java and C clients are making a mistake. But I also agree we should not deviate from them because most knowledgable ZK folks tend to be experts in ZK who switch-hit in polyglot environments so it's easier for the ecosystem if the behavior stays consistent across clients, even if it's arguably slightly incorrect (but clearly workable otherwise it would have been fixed long ago).

Right.  But I don't know if I would say "incorrect"; perhaps just a bit pessimistic and/or overly disruptive.

(Which is not out of character; ZK currently is of a pessimistic nature.  It would be nice to have something akin to [ZOOKEEPER-22](https://issues.apache.org/jira/browse/ZOOKEEPER-22), but that proposal has been lingering for a long time now!  Raft has that feature IIRC.)

> So I think we should do as you suggest:
> 
> 1. open a second PR with just the first fix

Done: https://github.com/python-zk/kazoo/pull/588.

That commit is based on the first fix, but with an expanded set of tests adapted from what I had cooked up for the second part.

> 2. keep the second commit in this PR around (ie, don't force-push and obliterate it), but close it for now so it doesn't get merged.

Okay; good idea.

> Furthermore, I think we should go one step further and submit an issue to the upstream ZK team to change this behavior in the Java and C clients.
> 
> Do you want to submit this or shall I? […] I just want to make sure in some way this is pushed upstream which which should result in either a fix or further clarification.

I am definitely going to keep this in mind, and on my TODO list.  I would like to think a bit more about which "use-cases" it could break, though.

(I could imagine an application batching asynchronous requests and hoping to get away with it by taking note of state changes; it could suddenly find itself in a strange place much more often.)

I am planning to ask for clarification on the ML a bit later, unless I identify an issue with this "proposal" in the meantime.  (I will report here in any case.)

But feel free to go ahead if you are curious and/or impatient, I'm not trying to collect points, and will gladly follow that conversation!

> As a heads up, my personal experience with the core ZK project is that it tends to be a slower moving project.

Right; as @ceache mentions, I have already gotten a taste of it :)   The good news is that they seem to have picked up some steam lately!

Cheers, -D

jeffwidman:
Sounds good. 

TBH, I switched teams at my day job last fall and am not currently responsible for any production ZK ensembles, so due to lack of time I unsubscribed from the mailing list... I pitch in here from time to time for fun and to give back to the community so that PR's don't languish. That said, I'll be curious to hear what the eventual outcome is.

PR #584, opened 2020-02-09
fix(tests): specify -Dfile.encoding=UTF-8 when starting the ZK JVM

ztzg:
`make_digest_acl_credential` assumes `UTF-8`, but ZooKeeper decodes digest auth packets using the JVM's default "charset"—which depends on the environment (e.g., `ISO-8859-1` if no locale is present).

This patch defines the system property which makes the JVM hosting ZooKeeper default to `UTF-8`, independently of the locale.

### DIFF
--- a/kazoo/security.py
+++ b/kazoo/security.py
@@ -59,7 +59,17 @@ READ_ACL_UNSAFE = [ACL(Permissions.READ, ANYONE_ID_UNSAFE)]
 
 
 def make_digest_acl_credential(username, password):
-    """Create a SHA1 digest credential"""
+    """Create a SHA1 digest credential.
+
+    .. note::
+
+        This function uses UTF-8 to encode non-ASCII codepoints,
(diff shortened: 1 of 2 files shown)


ztzg:
> I'm thinking it'd be good to also mention this `utf-8` assumption in the docstring of `make_digest_acl_credential` to aid debugging when users are like "this isn't working for me":
> Could you add that and then I'll merge?

Done; let me know if that's good enough.
 
> Also, I'm assuming:
> 
> * there's no way for the client to query the server to find the "charset"?

Not without modifying the ZooKeeper protocol, no.

> * that modifying `make_digest_acl_credential` to accept a charset param is more than a simple one-liner? In that case it's probably more trouble than it's worth and if someone else wants it they can submit a PR...

It would not be too complicated to update `make_digest_acl_credential`, but its callers would have to propagate the parameter. Moreover, I suspect that most ZK deployments will run under an UTF-8 locale in practice—and I would certainly recommend doing so.

Cheers, -D

PR #586, opened 2020-02-13
feat(core): Add create2 support

ceache:
Add support for `Create2` opcode (create returns ZkStat along side the path after a successful node creation).

### DIFF
--- a/kazoo/client.py
+++ b/kazoo/client.py
@@ -31,6 +31,7 @@ from kazoo.protocol.serialization import (
     CheckVersion,
     CloseInstance,
     Create,
+    Create2,
     Delete,
     Exists,
     GetChildren,
@@ -68,8 +69,11 @@ from kazoo.recipe.watchers import ChildrenWatch, DataWatch
 string_types = six.string_types
 bytes_types = (six.binary_type,)
 
-CLOSED_STATES = (KeeperState.EXPIRED_SESSION, KeeperState.AUTH_FAILED,
-               KeeperState.CLOSED)
+CLOSED_STATES = (
+    KeeperState.EXPIRED_SESSION,
+    KeeperState.AUTH_FAILED,
+    KeeperState.CLOSED
+)
 ENVI_VERSION = re.compile(r'([\d\.]*).*', re.DOTALL)
 ENVI_VERSION_KEY = 'zookeeper.version'
 log = logging.getLogger(__name__)
@@ -856,7 +860,7 @@ class KazooClient(object):
         return self.sync_async(path).get()
 
     def create(self, path, value=b"", acl=None, ephemeral=False,
-               sequence=False, makepath=False):
+               sequence=False, makepath=False, include_data=False):
         """Create a node with the given value as its data. Optionally
         set an ACL on the node.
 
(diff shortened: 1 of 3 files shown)


ceache:
This is a WIP. Still missing tests.

Note that when using these calls against Zk 3.4.x it does a round trip and throws a "NotImplemented" exception (there are talks of backporting to 3.4.x branch).

And this is working up to adding Container nodes (which return stat info as well): 
https://zookeeper.apache.org/doc/r3.5.6/zookeeperProgrammers.html#Container+Nodes

ceache:
Cool. Will do from now on

ceache:
Test added.

kazoo/client.py
```diff
@@ -68,8 +69,11 @@
 string_types = six.string_types
 bytes_types = (six.binary_type,)
 
-CLOSED_STATES = (KeeperState.EXPIRED_SESSION, KeeperState.AUTH_FAILED,
-               KeeperState.CLOSED)
+CLOSED_STATES = (
+    KeeperState.EXPIRED_SESSION,
+    KeeperState.AUTH_FAILED,
+    KeeperState.CLOSED
```
jeffwidman: 👍

kazoo/tests/test_client.py
```diff
@@ -775,6 +775,19 @@ def test_create_exists(self):
         path = client.create("/1")
         self.assertRaises(NodeExistsError, client.create, path)
 
+    def test_create_stat(self):
+        if TRAVIS_ZK_VERSION:
+            version = TRAVIS_ZK_VERSION
```
jeffwidman: Why is this needed rather than just sticking with the `version = self.client.server_version()` for everything?

Also, a lot of projects have been moving away from Travis over the past two years, I would not be surprised if we move away at some point ourselves.

kazoo/tests/test_client.py
```diff
@@ -775,6 +775,19 @@ def test_create_exists(self):
...
+        else:
+            version = self.client.server_version()
+        if not version or version < (3, 5):
+            raise SkipTest("Must use Zookeeper 3.5 or above")
```
jeffwidman: I haven't looked deeply at this code, so this is more of a side-comment, but long term it'd probably be more convenient to use `pytest`'s `skipif` decorator for things like this. That's a much bigger conversation about migrating the test suite to `pytest` and this is probably fine as-is if this pattern is consistently used elsewhere in Kazoo.

kazoo/tests/test_client.py
ceache: I did not look into this one. I am not sure of the value of `TRAVIS_ZK_VERSION` vs `client.server_version()`, I simply duplicated what was present in other tests.

Let me dig a little bit...

kazoo/tests/test_client.py
StephenSorriaux: I agree.

kazoo/client.py
```diff
@@ -994,10 +1019,16 @@ def create_completion(result):
         do_create()
         return async_result
 
-    def _create_async_inner(self, path, value, acl, flags, trailing=False):
+    def _create_async_inner(self, path, value, acl, flags,
+                            trailing=False, include_data=False):
         async_result = self.handler.async_result()
+        if include_data:
```
StephenSorriaux: Side-comment, I would have used the one-liner `opcode = Create2 if include_data else Create`

ceache: Ack. Note that this is going to get complicated once I add container and ttl...

StephenSorriaux: Sure, your way is better then

kazoo/tests/test_client.py
ceache: Only question I have is would a decorate be able to access `client.server_version()` ? It would require the client to be connected, no?

In either case, I agree with the move to pytest. Created #589.

PR #588, opened 2020-02-17
fix(core): allow requests to be queued in CONNECTING state (#374) [v2]

ztzg:
With this patch, requests issued while the client is in the `CONNECTING` state get queued instead of raising a misleading `SessionExpiredError`.

This fixes https://github.com/python-zk/kazoo/issues/374, and brings Kazoo more in line with the Java and C clients.

See the `kazoo.client.KazooClient.state` documentation as well as these discussions for more details:

  * https://github.com/python-zk/kazoo/pull/570#issuecomment-554798550
  * https://github.com/python-zk/kazoo/pull/583#issuecomment-586422386

This "v2" is a respin of https://github.com/python-zk/kazoo/pull/583 which only includes the fix for https://github.com/python-zk/kazoo/issues/374.

A few notes:

 - [X] Fixes the issue, and passes existing/adapted/new tests;
 - [X] The aforementioned queue is unbounded, but that matches the Java/C clients;
 - [X] The queue is emptied if the client leaves the `CONNECTING` state;
 - [X] The change in behavior is mentioned in the client documentation.

(Feel free to suggest a better place for that documentation.  I haven't updated `CHANGES.md` as that seems to be done at release time.)

### DIFF
--- a/docs/api/client.rst
+++ b/docs/api/client.rst
@@ -30,6 +30,27 @@ Public API
             A :class:`~kazoo.protocol.states.KazooState` attribute indicating
             the current higher-level connection state.
 
+            .. note::
+
+                Up to version 2.6.1, requests could only be submitted
+                in the CONNECTED state.  Requests submitted while
+                SUSPENDED would immediately raise a
+                :exc:`~kazoo.exceptions.SessionExpiredError`.  This
+                was problematic, as sessions are usually recovered on
+                reconnect.
+
+                Kazoo now simply queues requests submitted in the
+                SUSPENDED state, expecting a recovery.  This matches
+                the behavior of the Java and C clients.
+
+                Requests submitted in a LOST state still fail
+                immediately with the corresponding exception.
+
+                See:
+
+                  * https://github.com/python-zk/kazoo/issues/374 and
+                  * https://github.com/python-zk/kazoo/pull/570
+
     .. autoclass:: TransactionRequest
         :members:
         :member-order: bysource
--- a/kazoo/client.py
+++ b/kazoo/client.py
@@ -584,11 +584,11 @@ class KazooClient(object):
                 "and wouldn't close after %s seconds" % timeout)
 
     def _call(self, request, async_object):
-        """Ensure there's an active connection and put the request in
-        the queue if there is.
+        """Ensure the client is in CONNECTED or SUSPENDED state and put the
+        request in the queue if it is.
 
         Returns False if the call short circuits due to AUTH_FAILED,
-        CLOSED, EXPIRED_SESSION or CONNECTING state.
+        CLOSED, or EXPIRED_SESSION state.
 
         """
 
@@ -599,8 +599,7 @@ class KazooClient(object):
             async_object.set_exception(ConnectionClosedError(
                 "Connection has been closed"))
             return False
-        elif self._state in (KeeperState.EXPIRED_SESSION,
-                             KeeperState.CONNECTING):
+        elif self._state == KeeperState.EXPIRED_SESSION:
             async_object.set_exception(SessionExpiredError())
             return False
 
--- a/kazoo/tests/test_client.py
+++ b/kazoo/tests/test_client.py
@@ -8,7 +8,7 @@ import unittest
 import mock
 from mock import patch
 from nose import SkipTest
-from nose.tools import eq_
+from nose.tools import eq_, ok_, assert_not_equal
 from nose.tools import raises
 
 from kazoo.testing import KazooTestCase
@@ -492,9 +492,6 @@ class TestClient(KazooTestCase):
         self.assertRaises(AuthFailedError, client.create,
                           '/closedpath', b'bar')
 
-        client._state = KeeperState.CONNECTING
-        self.assertRaises(SessionExpiredError, client.create,
-                          '/closedpath', b'bar')
         client.stop()
         client.close()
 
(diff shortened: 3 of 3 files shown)


kazoo/tests/test_connection.py
```diff
@@ -100,32 +99,126 @@ def back(state):
 
     def test_connection_write_timeout(self):
         client = self.client
-        ev = threading.Event()
+        ev_suspended = threading.Event()
```
ceache: should `threading.Event` be `client.handler.event_object` instead to be gevent compatible?

kazoo/tests/test_connection.py
```diff
@@ -100,32 +99,126 @@ def back(state):
...
+        ev_can_connect = threading.Event()
+        ev_connected = threading.Event()
         path = "/" + uuid.uuid4().hex
         handler = client.handler
         _select = handler.select
-        _socket = client._connection._socket
+        _create_connection = handler.create_connection
+
+        def controlled_create_connection(*args, **kwargs):
+            ev_can_connect.wait(5)
```
ceache: This feels like a really involved test setup. Would you not be able to reproduce the same with
 * ZkClient configured with infinite retries, flat backoff, zero jitter
 * ZkServer initially downed.

1. You start the client -> it enters the connecting state as it keeps trying to connect
2. You can submit commands -> no exceptions (with your patch)
3. You start the server -> the commands should be effected.

I think it would make the test case less intrusive.
What do you think?

kazoo/tests/test_connection.py
ztzg: I think it should.  (I am acutely aware of that issue, as I made the mistake of copying a `time.sleep` statement from another test in a previous iteration—which deadlocked Travis :)

But: I did not want to introduce changes beyond what was necessary in this patch, and so have copied the "flow" from other tests.

What do you think?

@jeffwidman, @StephenSorriaux: Would you accept a subsequent "cleanup" patch for this?  If so, should it cover *all* the tests?  (That's going to be a bit noisy.)

kazoo/tests/test_connection.py
ztzg: Argh :)  What you are suggesting is close to what the previous test was doing:

https://github.com/python-zk/kazoo/pull/583/files#diff-b526bcbf3d9f758b2032e6763cb453c7R601

(Except for the "starting with ZK down" bit, which would indeed make things easier.  I don't know if any other tests start/stop ZK, though; I will check.)

I have switched because `test_connection_write_timeout`, which was already using a pretty intrusive method, was close to what I wanted to test—and I thought that making a single test a bit more complicated was the lesser evil.

The new `test_connection_lost_empties_queue` test, which is largely a duplicate of `test_connection_write_timeout`, also verifies that the queue empties out when the session cannot be recovered.  Would you know of a trick to invalidate the session without being intrusive?  (Besides submitting `_SESSION_EXPIRED`, which is inapplicable because it goes through the queue.)

I must admit that the thing which bothers me more in this test file is the duplication, which I made even worse; I was thinking of factoring out at least some of these local functions, and to submit a subsequent PR if the result turned out to be cleaner.

ceache: Hum, how about starting the client with a random ID, of course unknown to the server. Wouldn't that trigger a session expired error and clear the queues?

As for server up/down, I *think* I saw this in the tests pertaining to failovers. I might be wrong, would need a minute to go through the tests with a PC

kazoo/tests/test_connection.py
StephenSorriaux: You made a good point. I would say this PR should only cover what is necessary for the patch, I would be fine with keeping what was currently present and open an issue (& PR) about updating all tests in order to better test gevent and other libs.

ztzg: Okay; noted!

kazoo/tests/test_connection.py
ztzg: Okay—I have now moved by tests back to `TestClient`, and applied some of your suggestions:

Starting with ZK down is not easy, due to the way `ZookeeperCluster` the test suite interact, and the fact that `teardown_zookeeper` does not tear down the cluster :)  Moreover, the first state change a client perceives is `CONNECTED`; it never goes from the initial `LOST` to `SUSPENDED`.

So instead, I create a fresh client which is backed by a single server, and stop the latter to force a transition to `SUSPENDED`.

As for flushing: fudging the session password before reconnect is indeed a good way to trigger `SessionExpiredError`. (Surprisingly enough, just invoking `KazooClient._reset_session` isn't: a new session gets created, but the state directly transitions from `SUSPENDED` to `CONNECTED`!)

Moving the two tests back to `TestClient` means that they are also exercised under the `gevent` and `eventlet` handlers.

ztzg:
Hi @jeffwidman, @StephenSorriaux: Mind if I nag you a bit about this? :)

ztzg:
Fantastic!  Thank you.

Now, if I may take this opportunity to bother you a bit more... :) Do you have any plans regarding cutting a new release? And in any case: is there anything I could do which would help?

jeffwidman:
I think @StephenSorriaux is planning to cut a new release soon. 

Thank you again for your hard work on this! I had a deadline at work that kept me busy the past few weeks so had to defer open source involvement for a little bit.

ztzg:
Great!  (And no worries; I understand that this is not your full-time occupation :)

PR #590, opened 2020-02-23
feat(core): Add support for Create2 in transactions.

ceache:
Add `Create2` support in transactions.

Modeled after https://github.com/apache/zookeeper/blob/branch-3.5.6/zookeeper-server/src/main/java/org/apache/zookeeper/MultiResponse.java

### DIFF
--- a/kazoo/client.py
+++ b/kazoo/client.py
@@ -1590,7 +1590,7 @@ class TransactionRequest(object):
         self.committed = False
 
     def create(self, path, value=b"", acl=None, ephemeral=False,
-               sequence=False):
+               sequence=False, include_data=False):
         """Add a create ZNode to the transaction. Takes the same
         arguments as :meth:`KazooClient.create`, with the exception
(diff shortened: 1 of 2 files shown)


ceache:
This does not work for now, the Zookeeper server responds to my `Create2` ops with regular `Create` responses.

I am a little baffled, digging further into the Java code...

jeffwidman:
I assume you doublechecked the exact request that you were sending over the wire to make sure it was what you expect? There's a zookeeper plugin for Wireshark, although I doubt it includes the `create2` request...

ceache:
I knew and I did and it seems so far I stumbled upon a bug.

Create2 request with correct opcode goes in, Create reply comes back.

I am, right now, mucking around with the zookeeper code base, trying to
trace it.

Cheers,

jeffwidman:
@ceache curious why you closed this? 

Did you ever identify the bug source?

PR #598, opened 2020-04-06
feat(test): Move to pytest

### LINKED ISSUE #589: Migrate the test suite to pytest
As discussed in #586, leverage, among other things, pytest convenient decorator to remove noise from test cases.

ceache:
Switch to pytest. Close #589 

Switch over all tests to pytest new clean style.
Remove all nosetest'isms.
Run gevent/eventlet tests in Python3 as well.
Update Zookeeper to latest releases.


kazoo/tests/test_client.py
```diff
@@ -650,140 +677,148 @@ def test_create_makepath_incompatible_acls(self):
...
 
         client.ensure_path("/1/2/3/4")
-        self.assertTrue(client.exists("/1/2/3/4"))
+        assert client.exists("/1/2/3/4")
 
+    @pytest.mark.skip("BUG: sync() call is not chroot'd")
     def test_sync(self):
         client = self.client
-        self.assertTrue(client.sync('/'), '/')
+        assert client.sync('/') == '/'
```
ceache, line 764: The origin test was broken.  Turns out `sync()` is not unchrooted.

ceache:
I realize the documentations also needs to be updated

kazoo/tests/test_cache.py
```diff
@@ -111,25 +111,25 @@ def test_start(self):
...
         self.make_cache()
-        self.cache.start()
+        with pytest.raises(KazooException):
+            self.cache.start()
 
-    @raises(KazooException)
     def test_start_closed(self):
         self.make_cache()
-        self.cache.start()
+        # self.cache.start()
```
ceache, line 126: Broken test, always triggered here and thus no different than above

kazoo/tests/test_client.py
```diff
@@ -309,23 +325,25 @@ def listener(state):
...
+        assert states[0] == KazooState.CONNECTED
 
     def test_invalid_listener(self):
-        self.assertRaises(ConfigurationError, self.client.add_listener, 15)
+        with pytest.raises(ConfigurationError):
+            self.client.add_listener(15)
 
     def test_listener_only_called_on_real_state_change(self):
         from kazoo.protocol.states import KazooState
-        self.assertTrue(self.client.state, KazooState.CONNECTED)
```
ceache, line 328: broken test, always "`True`"

kazoo/tests/test_exceptions.py
```diff
@@ -9,17 +10,18 @@ def _get(self):
 
     def test_backwards_alias(self):
         module = self._get()
-        self.assertTrue(getattr(module, 'NoNodeException'))
-        self.assertTrue(module.NoNodeException, module.NoNodeError)
```
ceache, line 13: broken test, always "`True`"

requirements_test.txt
```diff
@@ -1,5 +1,6 @@
-coverage==3.7.1
```
ceache, line 1: These are pretty outdated. What guideline should be followed to pick new versions?

kazoo/tests/test_cache.py
StephenSorriaux: Good catch

kazoo/tests/test_client.py
StephenSorriaux: I am not sure to understand this one, what is the problem with this test?

requirements_test.txt
StephenSorriaux: Not sure about that. Maybe we can start from a clean sheet by setting the latest releases and accepting only fixes?

setup.cfg
```diff
@@ -11,6 +11,9 @@ tag_build = dev
 [metadata]
 license_file = LICENSE
 
+[tool:pytest]
+addopts = -ra -v
+
 [nosetests]
```
StephenSorriaux, line 17: Is this still needed?

requirements_test.txt
```diff
@@ -1,5 +1,6 @@
...
-mock==1.0.1
-nose==1.3.3
-flake8==2.3.0
-objgraph==3.4.0
+coverage
```
StephenSorriaux, line 1: Is this still needed since pytest-cov also have this dep?

kazoo/tests/test_client.py
ceache: Sync return the *unchrooted* path. So the test fails

requirements_test.txt
jeffwidman: Keep in mind these are pinned so that if someone is working on something unrelated, it doesn't start failing all their tests.

So now is a perfect time to bump them... I suggest bumping them all to the latest release and see if it passes. If it does, then we go with it. If not, then either fix or the pin to an older version that works, and file a ticket to update it.

kazoo/tests/test_queue.py
```diff
@@ -1,43 +1,46 @@
...
 from kazoo.tests.util import TRAVIS_ZK_VERSION
 
 
 class KazooQueueTests(KazooTestCase):
-
     def _makeOne(self):
         path = "/" + uuid.uuid4().hex
         return self.client.Queue(path)
 
     def test_queue_validation(self):
```
jeffwidman, line 14: This could be done as a pytest parametrized test... with "input" / "output exception"...

kazoo/tests/test_queue.py
```diff
@@ -66,88 +68,99 @@ def setUp(self):
...
                 skip = True
         if skip:
-            raise SkipTest("Must use Zookeeper 3.4 or above")
+            pytest.skip("Must use Zookeeper 3.4 or above")
 
     def _makeOne(self):
         path = "/" + uuid.uuid4().hex
         return self.client.LockingQueue(path)
 
     def test_queue_validation(self):
```
jeffwidman, line 77: this could also be changed to a pytest parametrized test

kazoo/tests/test_watchers.py
```diff
@@ -516,9 +517,10 @@ def changed(children):
             if counter > 0:
                 raise Exception("oops")
 
-        raises(Exception)(changed)
         counter += 1
-        self.client.create(self.path + '/' + 'smith')
+        with pytest.raises(Exception):
```
jeffwidman, line 521: Can this be changed to a more specific exception?

requirements_test.txt
ceache: I have split up requirements and version constraints. 
So our `requirement_*.txt` files are what we need or what we are (in)compatible with whereas the `constraints.txt` file contains the versions we pin.

Makes sense?

setup.cfg
ceache: removed

kazoo/tests/test_watchers.py
ceache: Done

requirements_test.txt
jeffwidman: After thinking about it more, I'd probably keep them in one file for simplicity, and then actually _not_ do any pinning of test requirements unless we _know_ a particular version's latest release is broken for us.

The worst that happens is a PR starts failing tests until we realize the issue, so it's not that big of a deal. KISS and all that.

And if someone says "hey, this version is broken" but it's not latest, then they can upgrade... unlike production deps, it's not hard to do for test-only deps.

Thoughts?

kazoo/tests/test_queue.py
ceache: Not sure what you mean by that. Can you show me an example? A link to some tests leveraging that?

requirements_test.txt
ceache: removed

kazoo/tests/test_queue.py
jeffwidman: https://docs.pytest.org/en/latest/parametrize.html#pytest-mark-parametrize-parametrizing-test-functions

```
@pytest.mark.parametrize("input,expected", [
    ({}, TypeError),
    (b"one", b"100", TypeError),
    (etc, etc),
])
def test_queue_validation(self, input, expected):
queue = self._makeOne()
with pytest.raises(expected):
    queue.put(input)
```

kazoo/tests/test_watchers.py
ceache: Good catch, it caught some more errors. `handler.sleep()` -> `handler.sleep_func()`

kazoo/tests/test_queue.py
ceache: I really struggled to get this to work with classes.
After much digging, it looks like this is an example of how it would work:
https://doc.pytest.org/en/3.0.5/example/parametrize.html#a-quick-port-of-testscenarios

Now, all that metaclass overhead really muddies the water in my opinion. Do you really want me to go down that route?
Alternatively, did I miss something obvious trying to get `pytest.mark.parametrize()` to work with `unittest.TestCase`?

kazoo/tests/test_watchers.py
```diff
@@ -216,30 +215,18 @@ def changed(children):
...
-
-    def test_bad_watch_func2(self):
-        counter = 0
-
-        @self.client.DataWatch(self.path)
-        def changed(d, stat):
-            if counter > 0:
-                raise Exception("oops")
-
-        raises(Exception)(changed)
```
ceache, line 230: This test was not doing anything (this expression evaluates to a function object, which is never called).
The test itself looks dubious to me. It tries to raise an exception in the watch, but that will be raised in the thread(let) handling watches, not the main client object context.
There are other tests for the watchers callback. So I think these can be removed.

kazoo/tests/test_watchers.py
```diff
@@ -490,35 +477,23 @@ def changed(children):
...
         update.wait(4)
-        eq_(update.is_set(), False)
-        eq_(all_children, ['smith'])
+        assert update.is_set() is False
+        assert all_children == ['smith']
 
         children = self.client.get_children(self.path)
-        eq_(sorted(children), ['george', 'smith'])
-
-    def test_bad_children_watch_func(self):
```
ceache, line 511: ditto

kazoo/tests/test_watchers.py
jeffwidman: Sounds good to me!

kazoo/tests/test_queue.py
jeffwidman: Ah, I forgot that these use `unittest.TestCase`... yeah, probably best to skip it for now. Perhaps at some point in the future we can drop `unittest` completely in favor `pytest` fixtures, as I find them a lot more powerful but that's beyond the scope of this PR... rather get this across the finish line.

jeffwidman:
Is this good to go or still working on it?

jeffwidman:
Congrats @ceache huge contribution!

Issue #602, opened 2020-04-12
User-defined authentication class cannot retry after the session expires and reconnects


QuantumEnergyE:
Because the scheme name of the custom authentication class is not digest, retry will not be triggered when the session expires and reconnects.

https://github.com/python-zk/kazoo/blob/34cdd60f26ff790618244450fd648b77f9f1353c/kazoo/protocol/connection.py#L726

ceache:
I believe this issue was already resolved in the 2.7.0 release.
Can you please confirm you do not see this issue when using the latest kazoo?

PR #603, opened 2020-04-12
Solve the problem that the user-defined authentication class cannot r…

QuantumEnergyE:
Support user-defined authentication class to retry after session expiration and reconnection.

### DIFF
--- a/kazoo/protocol/connection.py
+++ b/kazoo/protocol/connection.py
@@ -723,7 +723,7 @@ class ConnectionHandler(object):
         else:
             self._set_connected_ro_or_rw(client)
             for scheme, auth in client_auth_data_copy:
-                if scheme == "digest":
+                if scheme != "sasl":
                     ap = Auth(0, scheme, auth)
                     zxid = self._invoke(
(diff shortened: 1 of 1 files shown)


QuantumEnergyE:
https://github.com/python-zk/kazoo/issues/602

kazoo/protocol/connection.py
```diff
@@ -723,7 +723,7 @@ def _connect(self, host, port):
         else:
             self._set_connected_ro_or_rw(client)
             for scheme, auth in client_auth_data_copy:
-                if scheme == "digest":
+                if scheme != "sasl":
```
ceache, line 726: I am not sure I understand what you are trying to achieve here.
There is no such thing as a "sasl" scheme for client auth data. The SASL negotiation is done outside of Zookeeper client auth data

jeffwidman:
I see this is opened against the `release/2.6` branch... typically we aren't cutting bigfix releases against minor releases as we'd rather spend our limited volunteer time moving the project forward... please upgrade to 2.7 and if you still see an issue than open a PR against `master`.

Issue #607, opened 2020-05-01
Python 3.8 support


romainr:
I could not find an issue about it.

`async` is a reserved keyword now and will conflict, e.g. https://docs.python.org/3/reference/compound_stmts.html#async-def

```
python3.8/site-packages/kazoo/recipe/partitioner.py", line 185
    self._child_watching(self._allocate_transition, async=True)
                                                    ^
SyntaxError: invalid syntax
```

romainr:
For the record, I was on kazoo 2.0 and 2.7 compiles without reserved keywords.

I did not test the functionalities at run time yet.

StephenSorriaux:
Hello,

Kazoo officially supports Python 3.8. The current Kazoo version (2.7.0) would display a warning that was fixed in #609 and it will be integrated in the next Kazoo release.

romainr:
Great to know!

Adding 3.8 to https://pypi.org/project/kazoo/ would document it well

PR #611, opened 2020-05-09
feat(tests): add Python 3.8 and remove Python 3.6 from the testing suite

StephenSorriaux:
Relates to https://github.com/python-zk/kazoo/issues/607

## Changes

- Test Kazoo with CPython 2.7, CPython 3.7 and CPython 3.8.
- Deploy new Kazoo version only on CPython 3.8 job success.


ceache:
LGTM

PR #614, opened 2020-06-17
Use chain.from_iterable in threading.py

cool-RR:
This is a faster and more idiomatic way of using `itertools.chain`. Instead of computing all the items in the iterable and storing them in memory, they are computed one-by-one and never stored as a huge list. This can save on both runtime and memory space.

### DIFF
--- a/kazoo/handlers/threading.py
+++ b/kazoo/handlers/threading.py
@@ -178,7 +178,7 @@ class SequentialThreadingHandler(object):
         # anything to minimize changes
         if _HAS_EPOLL:
             # if the highest fd we've seen is > 1023
-            if max(map(_to_fileno, chain(*args[:3]))) > 1023:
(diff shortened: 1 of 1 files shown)


ceache:
Hi. The change looks good to me but would you mind amending your commit to follow our standard format?

ceache:
Specifically https://github.com/python-zk/kazoo/blob/master/CONTRIBUTING.md#git-commit-guidelines

cool-RR:
Gladly. The only thing missing is to prepend the commit message with "perf: ", correct?

ceache:
almost, the commit message should be:

```
perf(core): Use chain.from_iterable in threading.py

This is a faster and more idiomatic way of using itertools.chain.
Instead of computing all the items in the iterable and storing them
in memory, they are computed one-by-one and never stored as a huge
list. This can save on both runtime and memory space.
```

cool-RR:
Done.

PR #619, opened 2020-06-22
Handle SSL_WANT_READ/WRITE errors

### LINKED ISSUE #618: ConnectionLoss when reading large data over SSL
Using kazoo 2.7.0 or git master, I observe the following error
when calling get_children on a node with 8280 children when using
TLS:

```
  File "../zktest.py", line 17, in <module>
    zk.get_children('/nodepool/requests-lock')
  File "/home/corvus/kazoo/kazoo/client.py", line 1219, in get_children
    include_data=include_data).get()
  File "/home/corvus/kazoo/kazoo/handlers/utils.py", line 75, in get
    raise self._exception
kazoo.exceptions.ConnectionLoss
```

By removing the exception masking in _socket_error_handling I'm
able to see the underlying exception:

```
  File "/home/corvus/kazoo/kazoo/protocol/connection.py", line 605, in _connect_attempt
    response = self._read_socket(read_timeout)
  File "/home/corvus/kazoo/kazoo/protocol/connection.py", line 438, in _read_socket
    header, buffer, offset = self._read_header(read_timeout)
  File "/home/corvus/kazoo/kazoo/protocol/connection.py", line 226, in _read_header
    b = self._read(4, timeout)
  File "/home/corvus/kazoo/kazoo/protocol/connection.py", line 254, in _read
    chunk = self._socket.recv(remaining)
  File "/usr/lib/python3.5/ssl.py", line 914, in recv
    return self.read(buflen)
  File "/usr/lib/python3.5/ssl.py", line 791, in read
    return self._sslobj.read(len, buffer)
  File "/usr/lib/python3.5/ssl.py", line 577, in read
    v = self._sslobj.read(len)
ssl.SSLWantReadError: The operation did not complete (read) (_ssl.c:1981)
```

The [OpenSSL documentation](https://www.openssl.org/docs/manmaster/man3/SSL_get_error.html) indicates that when we receive this, we should call the function
again when the underlying socket is readable.  The same thing is
true for writing as well.

It is possible this bug is related to issues #587 and #580 as well.

jeblair:
This adds a simple recovery path in case an SSL connection receives
an SSL_WANT_READ or WRITE error.  Either error can occur while
reading or writing.  The error indicates that the underlying
operation should be retried after the socket is once again readable
or writable (per the error code).

Fixes #618

### DIFF
--- a/kazoo/protocol/connection.py
+++ b/kazoo/protocol/connection.py
@@ -6,6 +6,7 @@ import logging
 import random
 import select
 import socket
+import ssl
 import sys
 import time
 
@@ -247,7 +248,14 @@ class ConnectionHandler(object):
                         # have to check wlist and xlist as we don't set any
                         raise self.handler.timeout_exception(
                             "socket time-out during read")
-                chunk = self._socket.recv(remaining)
+                try:
+                    chunk = self._socket.recv(remaining)
+                except ssl.SSLError as e:
+                    if e.errno in (ssl.SSL_ERROR_WANT_READ,
+                                   ssl.SSL_ERROR_WANT_WRITE):
+                        continue
+                    else:
+                        raise
                 if chunk == b'':
                     raise ConnectionDropped('socket connection broken')
                 msgparts.append(chunk)
@@ -319,7 +327,14 @@ class ConnectionHandler(object):
                     raise self.handler.timeout_exception("socket time-out"
                                                          " during write")
                 msg_slice = buffer(msg, sent)
-                bytes_sent = self._socket.send(msg_slice)
+                try:
+                    bytes_sent = self._socket.send(msg_slice)
+                except ssl.SSLError as e:
+                    if e.errno in (ssl.SSL_ERROR_WANT_READ,
+                                   ssl.SSL_ERROR_WANT_WRITE):
+                        continue
+                    else:
+                        raise
                 if not bytes_sent:
                     raise ConnectionDropped('socket connection broken')
                 sent += bytes_sent


kazoo/protocol/connection.py
```diff
@@ -247,7 +248,14 @@ def _read(self, length, timeout):
                         # have to check wlist and xlist as we don't set any
                         raise self.handler.timeout_exception(
                             "socket time-out during read")
-                chunk = self._socket.recv(remaining)
+                try:
+                    chunk = self._socket.recv(remaining)
+                except ssl.SSLError as e:
+                    if e.errno in (ssl.SSL_ERROR_WANT_READ,
+                                   ssl.SSL_ERROR_WANT_WRITE):
```
ceache, line 255: Does it make sense to check for `WANT_WRITE` in the middle of a `recv()` call? Should the SSL try / catch not be aligned to the operation we are undertaking?

ceache:
Hi, thanks a lot for the issue report and the PR.

Do you mind amending your commit message to follow our recommendations?
Specifically https://github.com/python-zk/kazoo/blob/master/CONTRIBUTING.md#git-commit-guidelines

Secondly, as I understand it, this bug only tiggers on "large" operations (that do not fit in the SSL socket buffer, requiring a second read/write from the underlying socket).
Would you mind engineering a tests case for this? You mention creating 8k nodes would trigger the issue reliably.
This issue underscore the subtleties of SSL sockets that I would like permanently and regularly exercised in our test cases.

Thanks!

kazoo/protocol/connection.py
```diff
@@ -319,7 +327,14 @@ def _write(self, msg, timeout):
                     raise self.handler.timeout_exception("socket time-out"
                                                          " during write")
                 msg_slice = buffer(msg, sent)
-                bytes_sent = self._socket.send(msg_slice)
+                try:
+                    bytes_sent = self._socket.send(msg_slice)
+                except ssl.SSLError as e:
+                    if e.errno in (ssl.SSL_ERROR_WANT_READ,
```
ceache, line 333: ditto, does checking for `SSL_ERROR_WANT_READ` make sense for a `send()`?

kazoo/protocol/connection.py
jeblair: Yes, sorry I linked to the docs in the Issue but didn't here in the PR; this link should explain that due to the way OpenSSL's state machine works, either result can happen on either kind of underlying call: https://www.openssl.org/docs/manmaster/man3/SSL_get_error.html

ceache: Ok, thank you for clarifying. Makes sense

jeblair:
No problem on the commit message, sorry I didn't check the file first.

I'll look into adding a test.

jeblair:
Quick clarification: are there any existing SSL/TLS tests?

jeblair:
I didn't see any existing tests, so I added some commits to add an initial framework for SSL tests.  The fist upgrades the version of Zookeeper used by the tests to 3.6.1 since a newer version is required for SSL support at all.  The second updates the server to listen on an SSL port and adds a single simple client test over SSL.

Note that I have not yet written the requested test (8k children); I wanted to get this up before the end of my day.

jeblair:
I wrote the following test, and ran it with the ssl_want_read fix reverted, and it succeeded regardless (I also tried variations with more children):

```
    def test_many_children(self):
        client = self.client
        path = client.create("/test")
        for x in range(8200):
            path = client.create("/test/000000000000000000%s" % (x,))
        self.client.restart()
        client.get_children("/test")
```

While the error is consistently reproducible in our production environment, it may be more difficult to reproduce is other environments, such as tests.  Note that in particular, I ran my production test on a host that was remote from the Zookeeper server.  That introduces a considerable amount of delay (compared to a localhost connection) which may cause it more likely to hit the error.

I'd be happy to push up a commit with that test, but it may not be worth doing since it doesn't confirm the behavior.  I'll omit it for now, but let me know if you would rather me add it anyway.

ceache:
Thanks you @jeblair , it seems I sent you down a much bigger path than I expected.
I think your explanation of why you cannot reproduce the error makes perfect sense and, I don't see how you can reproduce this with out current test harness. Thank you very much for your contribution.

Code wise, things look good to me (other than the minor missing pytest module.
My only uncertainty with your PR lies in the sslfixture module: I don't know how the other @python-zk/maintainers feel about having certificates stored in the repo.

I'll let them comment.

jeblair:
No problem -- I think adding more SSL tests is great and I'm happy to help.  :)

Given that we're unlikely to be able to exercise this change in unit tests, and adding the SSL tests may need a bit more work (especially with respect to Zookeeper versions, certificates in tests, etc), I went a head and split that into a new PR.  So this one is now back to just the want_read/write fix.  The added SSL testing is in #620.

PR #620, opened 2020-06-26
test: add SSL test

jeblair:
This adds a simple SSL test along with the framework for running
the test Zookeeper in a mode where it listens on both SSL and
non-SSL ports.

This is based on earlier work in #619.

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -52,7 +52,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.7", "3.8", "3.9", "3.10", "pypy-3.7"]
+        python-version: ["3.7", "3.8", "3.9", "3.10", "pypy-3.10-v7.3.15"]
         zk-version: ["3.4.14", "3.5.10", "3.6.3", "3.7.1"]
         include:
           - python-version: "3.7"
@@ -63,7 +63,7 @@ jobs:
             tox-env: py39
           - python-version: "3.10"
             tox-env: py310
-          - python-version: "pypy-3.7"
+          - python-version: "pypy-3.10-v7.3.15"
             tox-env: pypy3
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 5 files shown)


kazoo/testing/sslfixtures.py
```diff
@@ -0,0 +1,138 @@
+ca_cert = """
```
StephenSorriaux, line 1: Would it be possible to generate all those certificates in a dynamic way (using pyOpenSSL for instance)?

jeffwidman:
nudge @jeblair

StephenSorriaux:
I rebased this PR but it is not working because ZK is expecting a keystore and a truststore (java stuff) instead of .pem files. Will try to get this working soon and add some "cert auto generation".

StephenSorriaux (approved):

StephenSorriaux:
@jeffwidman @ceache @jeblair this one is finalized, would love a review to get it merged.

kazoo/testing/__init__.py
```diff
@@ -1,6 +1,5 @@
 from kazoo.testing.harness import KazooTestCase, KazooTestHarness
 
-
```
jeffwidman, line 3: Interesting, I thought this was supposed to be two lines after imports, I'm surprised hound didn't complain... Maybe I'm wrong though?

jeffwidman (dismissed):
Looks great 👍

kazoo/testing/__init__.py
StephenSorriaux: I think you are right, so I just re-added even if Hound seemed to be happy

jeffwidman (approved):

jeblair:
Thanks for updating this.  ZK can use a pem file that is just a concatenation of the cert and key as its "keystore" without any special jks functionality.  But since that seems to have bitrotted in this PR somehow, doing all the actual jks stuff is probably going to be more robust in the long run.  :)

StephenSorriaux:
Thank you again for the PR @jeblair and thank you for the updated information. I remember using a product that would "weirdly" accept this kind of .pem files as a "keystore" and you are certainly right that it was ZK. My comment about the reason why the tests were failing was probably wrong.

PR #627, opened 2020-10-09
fix(client.py): Prevent unnecessary retries

AkaiNoCat:
Fixes #

## Why is this needed?
  When max_retry(old parameters) is not assigned, it results in an infinite retry until time_out
## Proposed Changes
  - change1: kazoo.client.py:
    `
_RETRY_COMPAT_DEFAULTS = dict(
    max_retries=0,
    retry_delay=0.1,
    retry_backoff=2,
    retry_max_delay=3600,
)
`
  Reduce the times of hidden and unnecessary retries

### DIFF
--- a/kazoo/client.py
+++ b/kazoo/client.py
@@ -80,7 +80,7 @@ log = logging.getLogger(__name__)
 
 
 _RETRY_COMPAT_DEFAULTS = dict(
-    max_retries=None,
+    max_retries=0,
     retry_delay=0.1,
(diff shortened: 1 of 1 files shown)


ceache:
This change has the down side that if that *one* connection attempt fails, it will not retry no matter the time out you set. 

The thought process is that it is more approachable to "try to connect for X time" then "try to connect that many times", especially when you factor in the number of servers in the ensemble etc

PR #631, opened 2020-12-08
Cannot resolve .. [Errno -2] Name or service not known

ksarabu1:
Fixes #

## Why is this needed?

Zookeeper is running from Docker containers and the service being resolved using Consul. When there's a network issue, the call to `socket.getaddrinfo` in method `_expand_client_hosts` causing it to return the error `Cannot resolve .. [Errno -2] Name or service not known`. This is causing retry to throwback `STOP_CONNECTING` and it stops retrying as result.

## Proposed Changes

Change: Raise `ForceRetryError` if method `_expand_client_hosts` returns empty host/ports and `client.hosts` has only one host to resolve.

```
        # Check for an empty hostlist, indicating none resolved
        if len(host_ports) == 0:
            if len(self.client.hosts) == 1:
                raise ForceRetryError('Reconnecting')
            return STOP_CONNECTING
```

n/a

### DIFF
--- a/kazoo/protocol/connection.py
+++ b/kazoo/protocol/connection.py
@@ -543,6 +543,8 @@ class ConnectionHandler(object):
 
         # Check for an empty hostlist, indicating none resolved
         if len(host_ports) == 0:
+            if len(self.client.hosts) == 1:
+                raise ForceRetryError('Reconnecting')
             return STOP_CONNECTING
 
         for host, hostip, port in host_ports:
--- a/kazoo/protocol/connection.py
+++ b/kazoo/protocol/connection.py
@@ -543,9 +543,7 @@ class ConnectionHandler(object):
 
         # Check for an empty hostlist, indicating none resolved
         if len(host_ports) == 0:
-            if len(self.client.hosts) == 1:
-                raise ForceRetryError('Reconnecting')
-            return STOP_CONNECTING
+            raise ForceRetryError('No host resolved. Reconnecting')
 
         for host, hostip, port in host_ports:
             if self.client._stopped.is_set():


kazoo/protocol/connection.py
```diff
@@ -543,6 +543,8 @@ def _connect_loop(self, retry):
 
         # Check for an empty hostlist, indicating none resolved
         if len(host_ports) == 0:
+            if len(self.client.hosts) == 1:
```
StephenSorriaux, line 546: Would you be willing to explain the reason for this additional condition? Looking at the code, I would have replaced the `return STOP_CONNECTING` with a `raise ForceRetryError('No host resolved, reconnecting')` but you may have seen something I am missing.

StephenSorriaux:
Thans for thir PR.

Overall I am fine with the idea behind the change, it makes sense to trigger the retry policy in case hosts are not resolved.

ceache:
I second @StephenSorriaux's feedback. The idea is good but I am confused by the additional check on a single host in the list.

kazoo/protocol/connection.py
ksarabu1: Thanks. I was just trying to keep the behavior to default when multiple hosts are used and if they all fail to resolve. I would also prefer to keep it simple and raise retry in all cases. Will update.

StephenSorriaux (approved):
Thank you.

ksarabu1:
@ceache  Can you please review.

ceache (approved):
Thanks!

ksarabu1:
Thank you. Can you please merge the PR? Thanks in advance.

PR #637, opened 2021-01-22
SequentialThreadingHandler support thread pool

SmartGeometry:
1. SequentialThreadingHandler support thread pool
2. DataWatch use thread pool version spawn in order to supper high
   frequency data update event

Fixes #

## Proposed Changes

  - change 1
  - change 2
  - ...

### DIFF
--- a/kazoo/handlers/threading.py
+++ b/kazoo/handlers/threading.py
@@ -20,6 +20,7 @@ import select
 import socket
 import threading
 import time
+from multiprocessing.pool import ThreadPool
 
 import six
 
@@ -112,6 +113,7 @@ class SequentialThreadingHandler(object):
         self._running = False
         self._state_change = threading.Lock()
         self._workers = []
+        self._pool = ThreadPool(30)
 
     @property
     def running(self):
@@ -280,6 +282,9 @@ class SequentialThreadingHandler(object):
         t.daemon = True
         t.start()
         return t
+
+    def safe_spawn(self, func, *args, **kwargs):
+        self._pool.apply_async(func, args, kwargs)
 
     def dispatch_callback(self, callback):
         """Dispatch to the callback object
--- a/kazoo/interfaces.py
+++ b/kazoo/interfaces.py
@@ -99,6 +99,17 @@ class IHandler(object):
 
         """
 
+    def safe_spawn(self, func, *args, **kwargs):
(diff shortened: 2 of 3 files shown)


SmartGeometry:
```
E                       OSError: [Errno 12] Cannot allocate memory
```

When building the jobs in pypy3, there was a memory error. I can't figure out the memory of the running environment.  Could you give me some advice, Thank you very much!

StephenSorriaux:
Hello,

Looking at the worker information, the running environment is a VM running in GCE infrastructure with, [considering this doc](https://docs.travis-ci.com/user/reference/overview/#virtualisation-environment-vs-operating-system), 7.5GB of RAM.

I tried to reproduce the issue locally, using a Ubuntu:Xenial container restricted to 6GB of RAM, with the exact same conditions, and did the tests with pypy3 using your branch and the current master. I ended up with:

- your branch is using a total up to 1.5GB of RAM and tests were OK. Although, I noticed the process running the tests has, at final, more than 10000 living threads. It is worth noting that the tests took 800s to execute.
```
CONTAINER ID   NAME            CPU %     MEM USAGE / LIMIT   MEM %     NET I/O          BLOCK I/O        PIDS
7c7459515c9b   loving_galois   126.75%   1.547GiB / 6GiB     25.79%    287MB / 1.67MB   19.4MB / 360MB   10126
```
```
root@7c7459515c9b:/# ps -huH -p 38707 | wc -l
10025
```
- current master is using a total up to 800MB of RAM. Threads did not cross the 100 threshold. Tests took 200s to execute
```
CONTAINER ID   NAME            CPU %     MEM USAGE / LIMIT   MEM %     NET I/O          BLOCK I/O        PIDS
7c7459515c9b   loving_galois   102.78%   735.5MiB / 6GiB     11.97%    287MB / 1.68MB   19.5MB / 405MB   159
```
```
root@7c7459515c9b:/# ps -huH -p 52285 | wc -l
65
```
At idle, the container is using around 400MB of RAM, so I think that the 7.5GB RAM announced by Travis is not real and it is more a 1GB of RAM VM instead. I am not that surprised as they changed their free tier recently and that we are still using travis-ci.org instead of travis-ci.com.

Although, it seems that the change you introduced also has a real impact on the number of thread generated. Was this expected?

PR #648, opened 2021-07-02
feat(recipe): add ExistingDataWatch class

jeblair:
This adds a subclass of DataWatch which only operates on existing
ZNodes.  If a user uses a DataWatch on a path and the ZNode at that
path is deleted, the DataWatch will still issue an "exists" call
and set a watch right before the final callback.  That means that
regardless of the return value of the callback and whether or not
Kazoo will invoke the callback again, the ZooKeeper server still
has a watch entry for that path.

In short, using a DataWatch on a path and then deleting that path
can leak watch entries on the ZooKeeper server.

Because the DataWatch recipe is designed to watch non-existing paths,
this behavior may be desired and relied on by some users, so it's
not considered a bug.  But other users may want to use DataWatches
for nodes where this behavior would be a problem.

The ExistingDataWatch class behaves similarly to its parent class,
DataWatch, but it does not set a watch on paths which do not exist
(whether that's because they never existed or were recently deleted).
This means that a user of an ExistingDataWatch can be assured that
after the callback with the deleted event, the watch is removed from
the server.

### DIFF
--- a/docs/api/recipe/watchers.rst
+++ b/docs/api/recipe/watchers.rst
@@ -15,6 +15,12 @@ Public API
 
         .. automethod:: __call__
 
+    .. autoclass:: ExistingDataWatch
+        :members:
+
+        .. automethod:: __init__
+
+        .. automethod:: __call__
 
     .. autoclass:: ChildrenWatch
         :members:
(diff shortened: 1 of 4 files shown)


jeffwidman (dismissed):
Makes sense to me. Thank you for including tests.

ceache:
Resurrecting this PR conversation a little bit.

I am just wondering in which case one would use `DataWatch` over `ExistingDataWatch`?
The leak issue rational is clear in the comment above, but what do you lose by removing that watch?

In short, I am wondering why we would want a different class instead of putting the fix in `Datawatch` proper.

Issue #649, opened 2021-07-30
Deadlock with read/write lock recipe


westphahl:
## Expected Behavior

Read/write lock recipe never deadlocks.

## Actual Behavior

Read/write lock recipe deadlocks under certain conditions. The deadlock seems to happen as follows:

1. **write lock** is acquired
2. another component tries to acquire the **read lock** and blocks (watching the **write lock** as predecessor)
3. yet another component tries to acquire the **write lock** and also blocks (watching the **read lock** as predecessor)
4. the first **write lock** is released which wakes up the **read lock**
5. the **read lock** starts watching the second **write lock** as predecessor :boom: 

We now have a deadlock with the read lock and the second write lock waiting for each other to be "released".

There can be a variation of this when the second write lock's znode is created before the read lock calls: https://github.com/python-zk/kazoo/blob/6337fd6f72b59fb20886f980f2e0d6d41525dc35/kazoo/recipe/lock.py#L263

This will lead to the same deadlock condition with the read and the second write lock watching each other independent of the currently acquired write lock.

## Snippet to Reproduce the Problem

```python
import logging
logging.basicConfig(format="%(asctime)s %(levelname)s:%(message)s", level=logging.DEBUG)

import threading
import time

from kazoo.client import KazooClient


k = KazooClient()
k.start()


lock_path = "/lock"
writer_running = threading.Event()
reader_running = threading.Event()


def writer():
    logging.info("Writer running")
    writer_running.set()
    with k.WriteLock(lock_path):
        logging.info("Got write lock")


def reader():
    logging.info("Reader running")
    reader_running.set()
    with k.ReadLock(lock_path):
        logging.info("Got read lock")


writer_thread = threading.Thread(target=writer)
reader_thread = threading.Thread(target=reader)

with k.WriteLock(lock_path):
    logging.info("Starting reader")
    reader_thread.start()
    reader_running.wait()
    logging.info("Starting writer")
    writer_thread.start()
    writer_running.wait()
    logging.info("Waiting for locks to block")
    time.sleep(5)

logging.info("Outer lock released")
logging.info("Joining reader/writer threads ...")
reader_thread.join()
writer_thread.join()
logging.error("This will never show up")
```

## Logs with logging in DEBUG mode

```
2021-07-30 12:06:08,586 INFO:Connecting to localhost(127.0.0.1):2281, use_ssl: True
[...]

PR #650, opened 2021-07-31
Fix possible deadlock in r/w lock recipe

westphahl:
The lock must only consider contenders with a sequence number lower than
it's own sequence number as also stated in the Zookeeper recipe
description for shared locks[0].

This wasn't working correctly as the ReadLock also considered WriteLocks
with a higher sequence number as contenders. This can lead to a deadlock
as described in #649.

[0]: https://zookeeper.apache.org/doc/r3.7.0/recipes.html#Shared+Locks

Fixes: #649


jeblair (approved):
Is it possible to add a test case for #649?

westphahl:
> Is it possible to add a test case for #649?

Done.

StephenSorriaux:
Thank you, I will first fix the CI since travis-ci.org has ceased and then take a look at your PR.

jeffwidman (commented):
This looks fine to me from a code perspective. I haven't used this lock recipe at all, so afraid I can't comment on the design change. But your explanation seems lucid/makes sense.

westphahl:
@StephenSorriaux I rebased the fix in order to have the CI jobs running.

StephenSorriaux:
Mh seems like Codecov has some trouble... Will retry the jobs tomorow

StephenSorriaux (approved):
Thanks a lot for this.

PR #654, opened 2021-10-18
feat(core): use gevent.select.poll to support fds > 1023

bellatoris:
When the highest fd in use is > 1023, route through gevent.select.poll.
Otherwise, use the existing select() behavior so by and large nothing
changes.


bellatoris:
> This is a duplicate of #656 and I think that PR's fixes the issue in a more holistic way (fixes all handlers).

Ok, I'll close this PR

PR #656, opened 2021-11-02
fix(core): use selectors to poll connections instead of raw select in threading,gevent,eventlet; solve the select limitation on a maximum file handler value and dynamic choose the best poller in the system

JetDrag:
Use selectors to poll connections instead of raw select in all kinds of handlers. To solve the limitation on a maximum file handler value and dynamic choose the best poller in the system.

## Why is this needed?

In the current implementation, the gevent, eventlet handler use the old select fuction which not support fd number langer than 1023 . Network Program with high numbers of opened sockets often raise `Exception: filedescriptor out of range in select()` 

## Proposed Changes

  - Use the `selectors` module which ship with official  python distro >=3.4 to simulate the select interface.
  - Change the select fuction in all handlers to use the `selectors` implementation.
  - The `selectors` will choose the best poller in the system.

### DIFF
--- a/kazoo/handlers/utils.py
+++ b/kazoo/handlers/utils.py
@@ -2,10 +2,20 @@
 
 import errno
 import functools
+import os
 import select
 import ssl
 import socket
 import time
+
+from collections import defaultdict
+
+import six
+
+if six.PY34:
+    import selectors
+else:
+    import selectors2 as selectors
 
 HAS_FNCTL = True
 try:
@@ -19,6 +29,7 @@ _NONE = object()
 
 class AsyncResult(object):
     """A one-time event that stores a value or an exception"""
+
     def __init__(self, handler, condition_factory, timeout_factory):
         self._handler = handler
         self._exception = _NONE
@@ -125,6 +136,7 @@ class AsyncResult(object):
                     functools.partial(callback, self))
             else:
                 functools.partial(callback, self)()
+
 
 def _set_fd_cloexec(fd):
     flags = fcntl.fcntl(fd, fcntl.F_GETFD)
@@ -272,6 +284,7 @@ def capture_exceptions(async_result):
     :param async_result: An async result implementing :class:`IAsyncResult`
 
     """
+
     def capture(function):
         @functools.wraps(function)
         def captured_function(*args, **kwargs):
@@ -279,7 +292,9 @@ def capture_exceptions(async_result):
                 return function(*args, **kwargs)
             except Exception as exc:
                 async_result.set_exception(exc)
+
         return captured_function
+
     return capture
 
 
@@ -291,6 +306,7 @@ def wrap(async_result):
     :param async_result: An async result implementing :class:`IAsyncResult`
 
     """
+
     def capture(function):
         @capture_exceptions(async_result)
         def captured_function(*args, **kwargs):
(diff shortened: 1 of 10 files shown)


kazoo/handlers/utils.py
```diff
@@ -291,12 +305,55 @@ def wrap(async_result):
...
+    selectors_module = selectors if not selectors_module else selectors_module
+    selector = selectors_module.DefaultSelector()
+    events_mapping = {selectors.EVENT_READ: rlist,
+                      selectors.EVENT_WRITE: wlist}
+    fd_events = defaultdict(int)
+    fd_fileobjs = defaultdict(list)
+
+    for event, fileobjs in events_mapping.items():
+        for fileobj in fileobjs:
+            fd_events[selectors._fileobj_to_fd(fileobj)] |= event
```
ceache, line 340: [nit] factor out the `_fileobj_to_fd()`

kazoo/handlers/utils.py
```diff
@@ -291,12 +305,55 @@ def wrap(async_result):
...
+def _selector_select(rlist, wlist, xlist, timeout=None, *,
+                     selectors_module=None):
+    """selector-based drop-in replacement for select to overcome select
+    limitation on a maximum filehandle value.
+
+    Need backport selectors2 package in python 2.
+    """
+    if timeout is None:
+        timeout = -1
+    selectors_module = selectors if not selectors_module else selectors_module
```
ceache, line 331: [nit] why not just have `selectors` as the default value for `selectors_module`?

kazoo/handlers/utils.py
```diff
@@ -291,12 +305,55 @@ def wrap(async_result):
...
             if value is not None:
                 async_result.set(value)
             return value
+
         return captured_function
+
     return capture
+
+
+def _selector_select(rlist, wlist, xlist, timeout=None, *,
```
ceache, line 322: You dropped `xlist` & `xevents` in this function. Was this intentional?

kazoo/tests/test_threading_handler.py
```diff
@@ -45,10 +45,6 @@ def test_double_start_stop(self):
         assert h._running is False
 
     def test_huge_file_descriptor(self):
-        from kazoo.handlers.threading import _HAS_EPOLL
```
ceache, line 48: I don't think this test has any remaining value with your change. This is really the responsibility of the various `selectors` module.

But I would like to see some unit tests for the   "`select()` API -> `selectors()` API -> `select()` result" conversions.

ceache (changes requested):
Thank you for your contribution.

I have add a couple of small nitpicks but otherwise this looks good.
My main code concerns are the `xlist` support and adding unit testing for the API conversion.

Can I also ask you to kindly follow our git commit naming guidelines https://github.com/python-zk/kazoo/blob/master/CONTRIBUTING.md#git-commit-guidelines ?

kazoo/handlers/utils.py
JetDrag: Because the selectors module didn't implement the support of xevent, the same as gevent/eventlet.

kazoo/tests/test_threading_handler.py
JetDrag: I just get the `select` test case code from official python test suite to test the `selector_select` and cover the normal `select` usage. See the file `kazoo/tests/test_selectors_select.py`

JetDrag:
Hi, @ceache I've reslove the all discussion through adding some fix and testcase. Please review again.

ceache (dismissed):
Thanks a lot for addressing my feedback. LGTM.

@StephenSorriaux are you good with this change too?

StephenSorriaux:
Hello, yes I am! Thanks a lot for this.

StephenSorriaux (dismissed):

JetDrag:
Hi, @StephenSorriaux @ceache  do you have time to merge the PR and release a new version? So we don't have to deploy a custom version in out production environment.

StephenSorriaux:
Hello @JetDrag, can you please rebase your branch with the current master so that we can check all tests are passing? Sorry for the delay, we were in the process of updating our CI.

ceache:
Hi, it seems to have failed

FAILED kazoo/tests/test_selectors_select.py::SelectTestCase::test_select_mutated

Cheers,

JetDrag:
@ceache sorry for the fail. I've update the test case, approval the flow again pls.

JetDrag:
@ceache The test cases passed in the official python version, but all greenlet related test case failed in the PyPy. It seems that the greenlet don't support the PyPy.

StephenSorriaux:
Can it be linked with the new `test_huge_file_descriptor` which creates lots of sockets that, if I am not wrong, are never closed?

JetDrag:
@StephenSorriaux I've updated the test cases and run in local PyPy env，I think you are correct. Pls approval the flow again .

StephenSorriaux (approved):

JetDrag:
It seems that the pr matched the condition to merge already. Could you merge the PR and release a new version, many thanks!

JetDrag:
Nice to see you in new year~ 

It would be a good choice to release a new version at the beginning of 2022 to fix this general problem. 
The modification of this PR has been validated in our production environment for months and reliable enough.

Cheers

PR #661, opened 2022-03-01
feat: allow setting the command buffer size

teebee:
Fixes https://github.com/python-zk/kazoo/issues/488

## Why is this needed?
Currently, the result of a Zookeeper command is truncated after 8KiB.

## Proposed Changes
https://github.com/python-zk/kazoo/issues/488#issuecomment-925782544

No.

### DIFF
--- a/kazoo/client.py
+++ b/kazoo/client.py
@@ -702,7 +702,7 @@ class KazooClient(object):
         """
         self._connection.close()
 
-    def command(self, cmd=b'ruok'):
+    def command(self, cmd=b'ruok', bufsize=8192):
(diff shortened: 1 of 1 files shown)


jeffwidman (commented):
Is it possible to write a test for this?

Could be relatively basic, such as setting to a very small custom value and passing a command known to be longer and then verifying the command doesn't work.

Similarly, could send a command longer than 8 KB and verify the default truncates it to 8kb... but that might be a little more painful to write, it's been a while since I've looked at the relevant test scaffolding on this repo.

PR #670, opened 2022-10-08
Backport changes from release/2.9

StephenSorriaux:
## Why is this needed?
- bump the version
- update changelog
- backport the fix that overrides the `egg_info` `tag-build` configuration

## Proposed Changes

  - bump version to 2.9.0
  - add changelog for 2.9.0
  - add some flags to `build` module to override `tag-build` value from `setup.cfg`

No


ceache (approved):

kazoo/version.py
```diff
@@ -1 +1 @@
-__version__ = '2.8.0'
```
ceache, line 1: FYI black, in my PR, will complain about the single quotes.

On a separate note, I think we should move to setuptools_scm and have this file generated for us from git tags, what do you think?

StephenSorriaux: Love the idea!

StephenSorriaux:
Mh I made a mistake by merging https://github.com/python-zk/kazoo/pull/671 before, because now I don't want to force push to the `release/2.9` branch to get those changes back. Will open another PR and cherry-pick.

kazoo/version.py
ceache: I have it ready to go, I'll submit a PR tomorrow. But I'd like it on top of my black PR if possible

PR #672, opened 2022-10-13
Catch socket timeout errors as well

excitoon:
## Why is this needed?

I feel this is a right thing. Let's discuss.


excitoon:
Opened by msitake.

excitoon:
Actual bug was pytest-dev/pytest#5502

PR #674, opened 2022-10-16
Backport changes from release/2.9

StephenSorriaux:
## Why is this needed?
- bump the version
- update changelog
- backport the fix that overrides the `egg_info` `tag-build` configuration

## Proposed Changes

  - bump version to 2.9.0
  - add changelog for 2.9.0
  - add some flags to `build` module to override `tag-build` value from `setup.cfg`

No


kazoo/version.py
```diff
@@ -1 +1 @@
-__version__ = '2.8.0'
+__version__ = '2.9.0'
```
ceache, line 1: Note my black PR will complain about the use of single quotes instead of double quotes. I'll rebase if need be. 

How do you feel about leveraging setuptool_scm to auto generate this file from git tag instead?

ceache (approved):

kazoo/version.py
StephenSorriaux: Thanks, I will merge this one then, and let you rebase yours.

Love the idea about the use of `setuptool_scm`. I will try to implement it in another PR, unless you already have something ready on your end!

PR #676, opened 2022-10-23
refactor: Use native Lock timeout instead of reimplementing

### LINKED ISSUE #605: Simplify lock recipe in a post Python 3.2 threading world.
The lock recipe assumes `threading.Lock.acquire()` does **not** take a timeout argument.

https://github.com/python-zk/kazoo/blob/b20c929421ff72421ad3770fb3436d583df5e5b0/kazoo/recipe/lock.py#L173

https://docs.python.org/3/library/threading.html#threading.Lock.acquire

> Changed in version 3.2: The timeout parameter is new.

Code can be simplified a lot now that Py2x and py3 < 3.6 support is dropped.

(Note that Gevent and Eventlet Lock already have a timeout argument)

a-ungurianu:
Fixes #605

## Why is this needed?

Not needed per se, but removes unnecessary complexity from the lock acquiring code

## Proposed Changes

  - use timeout parameter from the lock object instead of reimplementing timeout
  - rename `self._lock` to `self._thread_lock` (suggestions encouraged), as having a lock inside a lock made this hard to parse
  - ...

Nope

### DIFF
--- a/kazoo/recipe/lock.py
+++ b/kazoo/recipe/lock.py
@@ -134,7 +134,7 @@ class Lock(object):
         self._retry = KazooRetry(
             max_tries=None, sleep_func=client.handler.sleep_func
         )
-        self._lock = client.handler.lock_object()
+        self._acquire_method_lock = client.handler.lock_object()
 
     def _ensure_path(self):
         self.client.ensure_path(self.path)
@@ -174,27 +174,17 @@ class Lock(object):
             The ephemeral option.
         """
 
-        def _acquire_lock():
-            got_it = self._lock.acquire(False)
-            if not got_it:
-                raise ForceRetryError()
-            return True
-
         retry = self._retry.copy()
         retry.deadline = timeout
 
         # Ensure we are locked so that we avoid multiple threads in
         # this acquistion routine at the same time...
-        locked = self._lock.acquire(False)
-        if not locked and not blocking:
+        method_locked = self._acquire_method_lock.acquire(
+            blocking=blocking, timeout=timeout if timeout is not None else -1
+        )
+        if not method_locked:
             return False
-        if not locked:
-            # Lock acquire doesn't take a timeout, so simulate it...
-            # XXX: This is not true in Py3 >= 3.2
-            try:
-                locked = retry(_acquire_lock)
-            except RetryFailedError:
-                return False
+
         already_acquired = self.is_acquired
         try:
             gotten = False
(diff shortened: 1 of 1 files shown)


kazoo/recipe/lock.py
```diff
@@ -174,27 +174,17 @@ def acquire(self, blocking=True, timeout=None, ephemeral=True):
...
-
         retry = self._retry.copy()
         retry.deadline = timeout
 
         # Ensure we are locked so that we avoid multiple threads in
         # this acquistion routine at the same time...
-        locked = self._lock.acquire(False)
-        if not locked and not blocking:
+        thread_locked = self._thread_lock.acquire(
+            blocking=blocking, timeout=timeout if timeout is not None else -1
```
a-ungurianu, line 183: I'm not a fan that python's threading chose to require timeout to be a number. Thankfully, it seems the other handler implementation's underlying Lock replacement also follows the python API, so this should be safe.

I could just replace the default parameter to -1 instead. Don't have a strong opinion either way

a-ungurianu:
The test that's failing seems to suggest it's already flaky to begin with. Not sure how my change could affect that specific test. It seems that the client is failing to connect to all zookeeper instances, not just the ones we stopped.

StephenSorriaux:
Hey, thank you for this. Yes the tests around the "read only" feature is kind of flaky, mostly because one testcase is stopping the cluster. I have a hard time reproducing it locally so yeah...

kazoo/recipe/lock.py
StephenSorriaux: IMO the way you implemented it is better: it does not change the method signature, so no chance to break someone's current code.

StephenSorriaux (approved):
Thank you.

kazoo/recipe/lock.py
```diff
@@ -134,7 +134,7 @@ def __init__(self, client, path, identifier=None, extra_lock_patterns=()):
         self._retry = KazooRetry(
             max_tries=None, sleep_func=client.handler.sleep_func
         )
-        self._lock = client.handler.lock_object()
+        self._thread_lock = client.handler.lock_object()
```
ceache, line 137: You asked for a suggestion: how about `_handler_lock`?
Just to recognize that other handlers are not thread based

ceache:
LGTM
Thanks a lot for this.

a-ungurianu:
My pleasure. I'm trying to learn more about Zookeeper and the Kazoo recipes. Happy give back :)

kazoo/recipe/lock.py
a-ungurianu: Actually, how about "_acquire_method_lock" which describes what it does instead of what it is.

ceache:
Sure. Sounds good to me 👍

ceache (approved):

StephenSorriaux (approved):

PR #678, opened 2022-10-28
refactor: Remove vestiges of python2 support

### LINKED ISSUE #677: Remove usage of `six`
`six` and the things it provides are only needed for the transition from python2 to python3. As it seems that this library no longer supports python2, it'd be nice to remove the usage of this library.

Happy to help clean it up

a-ungurianu:
Fixes #677 

## Why is this needed?

Not "needed", but this removes unnecessary dependencies on `six` and `selectors2` making this package a no-deps package for runtime usage. 

More of a spring cleaning

## Proposed Changes

  - remove usage of `six`
  - remove ImportError conditional statements that were due to `python3` stdlib updates

Should be completely backwards compatible. Exceptions might change a bit as we use `python3`'s `raise ... from ...` syntax


kazoo/handlers/threading.py
```diff
@@ -17,16 +17,11 @@
...
-
 import kazoo.python2atexit as python2atexit
 from kazoo.handlers import utils
 from kazoo.handlers.utils import selector_select
 
-try:
-    import Queue
-except ImportError:  # pragma: nocover
-    import queue as Queue
+import queue as Queue
```
ceache, line 24: I would prefer you rename it `queue` instead of keeping the legacy capitalized name

kazoo/handlers/utils.py
```diff
@@ -10,18 +10,8 @@
...
 from collections import defaultdict
 
-import six
-
-if six.PY34:
-    import selectors
-else:
-    import selectors2 as selectors
-
-HAS_FNCTL = True
```
ceache, line 20: `fcntl` is not a py27 thing (it was present then too). It is a runtime thing.

kazoo/handlers/utils.py
```diff
@@ -358,26 +347,19 @@ def fileobj_to_fd(fileobj):
...
-
-    Need backport selectors2 package in python 2.
     """
     if timeout is not None:
-        if not (
-            isinstance(timeout, six.integer_types)
-            or isinstance(timeout, float)
-        ):
+        if not (isinstance(timeout, int) or isinstance(timeout, float)):
             raise TypeError("timeout must be a number")
```
ceache, line 356: [nit] `isinstance()` accepts a list as second argument

ceache (commented):
Looks good. Thanks for taking care of this

kazoo/handlers/utils.py
a-ungurianu: Interesting. The docs says that it's not supported only in WASM. Are there people who use this lib with WASM?
I agree with keeping it, just curious

ceache: Not that I know of....
It may simply be that much older versions of python did not have it.

kazoo/protocol/connection.py
```diff
@@ -808,23 +806,11 @@ def _authenticate_with_sasl(self, host, timeout):
             try:
                 response = sasl_cli.process(challenge=challenge)
             except puresasl.SASLError as err:
-                six.reraise(
-                    SASLException,
-                    SASLException("library error: %s" % err),
-                    sys.exc_info()[2],
-                )
+                raise SASLException("library error: %s" % err) from err
```
jeffwidman, line 809: Do we still need the `: %s" % err` in these three?

Would `raise SASLException("library error") from err` work?

kazoo/protocol/connection.py
```diff
@@ -838,13 +824,9 @@ def _authenticate_with_sasl(self, host, timeout):
...
                 header, buffer, offset = self._read_header(timeout)
-            except ConnectionDropped:
+            except ConnectionDropped as ex:
                 # Zookeeper simply drops connections with failed authentication
-                six.reraise(
-                    AuthFailedError,
-                    AuthFailedError("Connection dropped in SASL"),
-                    sys.exc_info()[2],
-                )
+                raise AuthFailedError("Connection dropped in SASL") from ex
```
jeffwidman, line 829: This is a bit inconsistent with the refactoring above that adds a `% err` to the error messages, but here is omitted.

I suspect this is the right format and the others should be updated, but if not then this should be tweaked for consistency.

jeffwidman:
Thanks for tackling this!

kazoo/protocol/connection.py
ceache: [nit] I would also recommend you keep using `err` for exceptions, to stay consistent with the others.

kazoo/recipe/lock.py
```diff
@@ -207,13 +200,12 @@ def _acquire_lock():
                 )
             except RetryFailedError:
                 pass
-            except KazooException:
+            except KazooException as ex:
```
ceache, line 203: [nit] ditto about the consistency of using `err` vs `ex`.

ceache (commented):
Couple minor nitpicks. LGTM otherwise

kazoo/protocol/connection.py
a-ungurianu:
```
>>> def throw():
...     raise KeyError("Message 1")
...
>>> def throw_through():
...     try:
...         throw()
...     except KeyError as err:
...         raise Exception("Message 2") from err
...
>>> throw_through()
  File "<stdin>", line 3, in throw_through
  File "<stdin>", line 2, in throw
KeyError: 'Message 1'

The above exception was the direct cause of the following exception:

  File "<stdin>", line 1, in <module>
  File "<stdin>", line 5, in throw_through
Exception: Message 2
```

This is what this would produce. Happy to drop it, just want to make sure we're happy with the "change" in exception messaging

kazoo/recipe/lock.py
a-ungurianu: There's not a lot of consistency overall. I was inferring the name from the class name, `...Exception as ex`, `...Error as err`.

What do you think?

ceache: I think that makes perfect sense 😄 

I did a quick poll of the codebase and it is not very consistent indeed.

We currently mostly have `err`, `exc` and (my "favorite") the ambivalent `e`.

   1 ex
   6 err
   6 exc
   7 e

I would suggest we then follow your model but with `*Error` / `*Exception` aligned with `err` / `exc` respectively.

We can remediate the others overtime.
Would you be OK with that?

kazoo/protocol/connection.py
ceache: I think this is OK.  Having the stringification of parent exception in the message does not provide any value.

We only expose "library", "protocol" and "unknown" messages at this point anyway. I think this is good 👍

kazoo/tests/conftest.py
```diff
@@ -4,7 +4,10 @@
...
-    log.error("Zookeeper cluster logs:")
-    for logs in cluster.get_logs():
-        log.error(logs)
+    if hasattr(node._testcase, "cluster"):
+        cluster = node._testcase.cluster
+        log.error("Zookeeper cluster logs:")
+        for logs in cluster.get_logs():
+            log.error(logs)
+    else:
+        log.error("Cluster was not up!")
```
StephenSorriaux, line 13: Some tests does not make use of the cluster, I think the `else` case can just be omitted here. What do you think?

kazoo/recipe/lock.py
```diff
@@ -197,13 +190,12 @@ def acquire(self, blocking=True, timeout=None, ephemeral=True):
...
                 pass
-            except KazooException:
+            except KazooException as exc:
                 # if we did ultimately fail, attempt to clean up
-                exc_info = sys.exc_info()
                 if not already_acquired:
                     self._best_effort_cleanup()
                     self.cancelled = False
-                six.reraise(exc_info[0], exc_info[1], exc_info[2])
+                raise exc
```
StephenSorriaux, line 198: Can't we just `raise` here since it seems we are not making use of `exc`?

a-ungurianu: That's very fair. Forgot you can do that.

kazoo/tests/conftest.py
a-ungurianu: I think you're right. The lack of a cluster attribute is a marker that the test was not meant to have a cluster.

I'll remove the else.

StephenSorriaux (approved):
Many thanks for your work on this!

StephenSorriaux:
Gentle ping @ceache and @jeffwidman, would you like to give it a last look?

kazoo/handlers/eventlet.py
```diff
@@ -1,6 +1,8 @@
 """A eventlet based handler."""
 from __future__ import absolute_import
 
+import atexit
+
```
jeffwidman, line 5: No need for blank line, these are all stdlib imports, right? So can be grouped together in lexicographic order. This comment applies multiple places in the PR.

kazoo/handlers/threading.py
```diff
@@ -12,21 +12,17 @@
...
-
-import kazoo.python2atexit as python2atexit
 from kazoo.handlers import utils
 from kazoo.handlers.utils import selector_select
 
-try:
-    import Queue
-except ImportError:  # pragma: nocover
-    import queue as Queue
+import queue
```
jeffwidman, line 25: Shouldn't this be grouped with the `stdlib` imports group above?

kazoo/protocol/connection.py
```diff
@@ -808,23 +806,11 @@ def _authenticate_with_sasl(self, host, timeout):
...
-                six.reraise(
-                    AuthFailedError,
-                    AuthFailedError("Unknown error: %s" % err),
-                    sys.exc_info()[2],
-                )
+                raise SASLException("library error") from err
+            except puresasl.SASLProtocolException as exc:
+                raise AuthFailedError("protocol error") from exc
+            except Exception as exc:
+                raise AuthFailedError("Unknown error") from exc
```
jeffwidman, line 813: I know this was already partially discussed, but why mix `err` and `exc`? Here it just adds confusion... pick one in this section for consistency...

kazoo/recipe/lock.py
```diff
@@ -15,16 +15,9 @@
 
 """
 import re
-import sys
-
-try:
-    from time import monotonic as now
-except ImportError:
-    from time import time as now
+from time import monotonic as now
```
jeffwidman, line 18: Why the rename? Why not `import monotonic` and use `monotonic` at the callsite? The original rename was because of the try/catch import, but here that's no longer relevant...

jeffwidman (commented):
A few minor nits. I may be off on them, and I'd also be okay if this shipped as-is, but if you want my $0.02 here it is 😀

kazoo/recipe/lock.py
ceache: I would further say that I much prefer importing modules instead of classes or functions (kind of like one would use namespaces in C++).

This is much more manageable because you avoid any name collision and it is very clear where each symbol comes from.

``` python
import time

def foo():
    tick = time.monotonic()

```

ceache (commented):
I have done one last pass.
Code is good, this is all styling feedback.

Thanks for all this!

a-ungurianu:
Any idea why isort take care of the import ordering/grouping?

kazoo/handlers/gevent.py
```diff
@@ -19,7 +21,6 @@
     from gevent.coros import Semaphore, RLock
```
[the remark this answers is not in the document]
a-ungurianu: I can tag it along as a separate commit

StephenSorriaux (approved):

StephenSorriaux:
@a-ungurianu Thank you again for this. I'll let you free to merge this whenever you want, and, if I am not mistaken, make it your first merge as an official Kazoo contributor :rocket:

ceache (approved):
Thanks for all of this 👍

a-ungurianu:
Didn't realize I can press the big green button :). Here goes nothing

jeffwidman:
Thanks again for driving this to completion! Congrats on your first merge.

Issue #679, opened 2022-11-03
Kazoo 2.9.0 throws "Connection dropped: socket connection error: The handle is invalid"


fafairuz:
Kazoo 2.9.0 throws `Connection dropped: socket connection error: The handle is invalid` and from server, I got `EndOfStreamException: Unable to read additional data from client, it probably closed the socket: address = /10.1.17.34:3664, session = 0x0`

I have tried with previous version (2.8.0) and it is working fine.

## Expected Behavior
KazooClient can be started after issuing `zk.start()`

## Actual Behavior
Kazoo returning `Connection dropped: socket connection error: The handle is invalid` after issuing `zk.start()`

## Snippet to Reproduce the Problem
```
Python 3.8.10 (tags/v3.8.10:3d8993a, May  3 2021, 11:48:03) [MSC v.1928 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> from kazoo.client import KazooClient
>>> zk_hosts    = "zookeeper_host_1:2181"
>>> zk = KazooClient(hosts=zk_hosts)
>>> zk.start()
Connection dropped: socket connection error: The handle is invalid
Connection dropped: socket connection error: The handle is invalid
Connection dropped: socket connection error: The handle is invalid
Connection dropped: socket connection error: The handle is invalid
Connection dropped: socket connection error: The handle is invalid
Connection dropped: socket connection error: The handle is invalid
Connection dropped: socket connection error: The handle is invalid
Connection dropped: socket connection error: The handle is invalid
Connection dropped: socket connection error: The handle is invalid
Failed connecting to Zookeeper within the connection retry policy.
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "C:\Python38\lib\site-packages\kazoo\client.py", line 635, in start
    raise self.handler.timeout_exception("Connection time-out")
kazoo.handlers.threading.KazooTimeoutError: Connection time-out
>>>
```
## Logs with logging in DEBUG mode

This is logs from zookeeper host. It is repeated numbers of times.
```
Nov 03 17:19:39 nsl-zookeeper-02 zkServer.sh[1452]: EndOfStreamException: Unable to read additional data from client, it probably closed the socket: address = /my_client_IP:4089, session = 0x0
[...]

PR #681, opened 2022-11-04
fix(core): fix for Windows platform

StephenSorriaux:
Fixes https://github.com/python-zk/kazoo/issues/679

## Why is this needed?
This fixes an issue with the Windows platform that appeared in the latest 2.9.0 release. This also adds a very basic testing for the Windows platform (only 1 job, for the latest version of Python and the latest version of ZK).

## Proposed Changes
  - remove the `os.fstat(fd)` check that is invalid on Windows platform
  - adjust `test_selector_select` test
  - add a single job to test Windows platform (`gevent` seems broken on Windows so... skipped it): only 1 job because I don't feel like we need to re-test every version of Python/ZK for Windows

Nah.

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -17,11 +17,9 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - name: Handle the code
-        uses: actions/checkout@v4
+      - uses: actions/checkout@v4
 
-      - name: "Set up Python 3.10"
-        uses: actions/setup-python@v4
+      - uses: actions/setup-python@v4
         with:
           python-version: "3.10"
 
@@ -47,9 +45,8 @@ jobs:
     needs: [validate]
 
     name: >
-      Test Python ${{ matrix.python-version }},
+      Linux - Test Python ${{ matrix.python-version }},
       ZK ${{ matrix.zk-version }}
-
     runs-on: ubuntu-latest
 
     strategy:
@@ -69,8 +66,7 @@ jobs:
           - python-version: "pypy-3.7"
             tox-env: pypy3
     steps:
-      - name: Handle the code
-        uses: actions/checkout@v4
+      - uses: actions/checkout@v4
 
       - name: Set up Python ${{ matrix.python-version }}
         uses: actions/setup-python@v4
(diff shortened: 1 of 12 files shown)


kazoo/handlers/utils.py
```diff
@@ -354,7 +354,11 @@ def fileobj_to_fd(fileobj):
             raise TypeError("Invalid file object: " "{!r}".format(fileobj))
     if fd < 0:
         raise TypeError("Invalid file descriptor: {}".format(fd))
-    os.fstat(fd)
+
+    # XXX: needed?
+    # selectors_module.register handle fileobject, do we need the use of fileobj_to_fd()?
+    # is this test valid for Windows?
+    # os.fstat(fd)
```
StephenSorriaux, line 361: Hello @JetDrag,

Sorry to summon you here. This `os.fstat(fd)` call is generating crashes on Windows because the handle is, it seems, invalid. With this call commented out, the crash does not appear (and tests are running fine).
Since you did this implementation, I was wondering if:
- there was a reason using the `fileobj_to_fd()` method instead of directly giving the fileobj to `selectors_module.register()` (which seems to be capable of handling fileobj directly)?
- you know if the `os.fstat(fd)` test is valid for Windows?

Thank you.

ceache: I second this.
This `os.fstat()` does not do anything. 
Actually, python3 handles `os.stat()` on FD or file just fine.

ceache: https://docs.python.org/3/library/os.html#os.stat

JetDrag: Hi, @StephenSorriaux @ceache 

The function `fileobj_to_fd` is same as `_fileobj_to_fd` function in `selectors` module  almost except the last second line `os.fstat(fd)`.

I did this implementation to simulate the raw select behavior via test case `test_selectors_select`, which is modified from Python official test cast `test_select`.

If I don't do that, the test cast `test_errno` in the `test_selectors_select` will fail, because the exception (OSError with errno 22, EINVAL) raise by  `selectors._BaseSelectorImpl` module don't meet the test case requirement：raise OSError with errno 9 (EBADF) when select meet a closed fd.

Maybe we cloud make some other simulate implementation to make those spec work on the Windows?

ceache: https://github.com/python/cpython/blob/main/Lib/test/test_select.py#L32  `test_errno()` for reference.

I do not think we care about this behavior in Kazoo.
Actually, the err conditions with selector when selecting "bad file descriptors" are varied depending on platform and selector implementations (there are threads about this on the Python bug report).
This is here trying to make sure that selectors implementations are all equivalente to `select()` behavior and we really can't do that.

IMHO. the only thing we need to do is ensure how Kazoo reacts to a closed FD.
In other words, be it `EINVAL` or `EBADF` or other, we should raise an uniform Kazoo error (with Python3, we could pass up the original, OS/Implementation dependent Exception).

What do you think?

ceache:
@StephenSorriaux  is this a draft or ready for review yet?

StephenSorriaux:
@ceache Not really, it is still a draft. I'm just too used to gitlab, which auto marks a PR as draft if "draft" is in the title.

kazoo/handlers/utils.py
JetDrag: It maybe the best solution.

StephenSorriaux: Thank you both for your thoughts. As suggested, I tried to unify the exceptions that can be raised and catch that in the `test_errno()` testcase. That may be too naive, not sure really, so I would love your feedback on this.

jeffwidman:
What's needed to move this forward? It'd be nice to get this resolved for our Windows users.

StephenSorriaux:
I was waiting for feedbacks from @JetDrag / @ceache but I will go ahead and just finish it so that it can be reviewed.

ceache (dismissed):
Thanks!

jeffwidman (approved):
Looks fine to me, with the caveat that I haven't actually written/run python on 🪟 in years.

JetDrag:
Excellent work！ @StephenSorriaux

.github/workflows/testing.yml
```diff
@@ -111,3 +110,54 @@ jobs:
 
       - name: Publish Codecov report
         uses: codecov/codecov-action@v3
+
+  test_windows:
+    needs: [validate]
+    name: Windows - Test Python 3.10, ZK 3.7.1
```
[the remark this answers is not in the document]
StephenSorriaux: Thanks, I updated some of the names

.github/workflows/testing.yml
```diff
@@ -111,3 +110,54 @@ jobs:
...
+    steps:
+      - name: Handle the code
+        uses: actions/checkout@v3
+
+      - name: "Set up Python 3.10"
+        uses: actions/setup-python@v4
+        with:
+          python-version: "3.10"
+
+      - name: Handle pip cache
```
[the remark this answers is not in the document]
StephenSorriaux: I think it could, but I am not sure this is something I want to try in this PR

StephenSorriaux:
@ceache @jeffwidman Sorry to resurrect this only now, would appreciate another review if possible and finally merge this.

jeffwidman (approved):
LGTM, with the caveat that I don't run Windows. But I think the additional confidence of the basic sanity test is enough to be comfortable merging this.

Thank you for your hard work on this!

PR #682, opened 2022-11-05
recipe(cache): Use protocol.paths.join instead of os.path.join

### LINKED ISSUE #680: kazoo.recipe.cache.TreeCache doesn't work on Windows because it uses backslash
the recipe is using os.path.join which uses backslash on windows. 
https://github.com/python-zk/kazoo/blob/master/kazoo/recipe/cache.py#L339
need to use '/'.join() or similar instead

## Expected Behavior
TreeCache should load the tree

## Actual Behavior
Cannot load any children because uses \ on windows
e.x. tries to get '/presence\\tss' instead of '/presence/tss'

## Snippet to Reproduce the Problem

Listen to any tree on windows 

## Logs with logging in DEBUG mode
DEBUG:kazoo.client:Sending request(xid=305): GetChildren(path='/presence\\tss', watcher=<bound method TreeNode._process_watch of <kazoo.recipe.cache.TreeNode object at 0x000001FD4BA1E5E0>>)

## Specifications

  - Kazoo version: 2.6+, but I think it is present in all versions
  - Result of `pip list` command: 
  - Zookeeper version: 2.6 on Linux
  - Zookeeper configuration: put here any useful ZK configuration (authentication, encryption, number of ZK members, number of (concurrent?) clients, Java version, krb5 version, etc.)
  - Python version: 3.8
  - OS: Windows

a-ungurianu:
Fixes #680 

## Why is this needed?

`os.path.join` doesWindows path joining on Windows resulting in bad zookeeper paths

## Proposed Changes

- use the existing path joining method in `kazoo.protocol.paths` instead of `os.path.join`.

No.

### DIFF
--- a/kazoo/recipe/cache.py
+++ b/kazoo/recipe/cache.py
@@ -16,12 +16,10 @@ import contextlib
 import functools
 import logging
 import operator
-import os
 
 from kazoo.exceptions import NoNodeError, KazooException
-from kazoo.protocol.paths import _prefix_root
+from kazoo.protocol.paths import _prefix_root, join as kazoo_join
 from kazoo.protocol.states import KazooState, EventType
-
 
 logger = logging.getLogger(__name__)
 
@@ -336,7 +334,7 @@ class TreeNode(object):
             if result.successful():
                 children = result.get()
                 for child in sorted(children):
-                    full_path = os.path.join(path, child)
+                    full_path = kazoo_join(path, child)
                     if child not in self._children:
                         node = TreeNode(self._tree, full_path, self)
                         self._children[child] = node


ceache (approved):
Looks good to me 👍
Thanks for taking care of this!

a-ungurianu:
> @a-ungurianu Do you mind rebasing instead of merging master? Makes for cleaner history

I agree. Fat fingered the GitHub UI (I think you can change the default in the repo settings somehow)

a-ungurianu:
Redid the rebase

StephenSorriaux (approved):
Thank you!

ceache (approved):
Thanks!

Issue #458, opened 2017-07-19
Deadlock with TreeCache and reconnection


nekto0n:
Hi there!
First of all, thank you for such a great recipe as `TreeCache`. We recently tried to use it, but started experiencing ussies upon reconnection with large trees. After reconnecting every request fails with ConnectionLoss error. I managed to capture a stacktrace:
```
  File "/venv/lib/python2.7/site-packages/gevent/greenlet.py", line 327, in run
    result = self._run(*self.args, **self.kwargs)
  File "/venv/lib/python2.7/site-packages/kazoo/protocol/connection.py", line 473, in zk_loop
    if retry(self._connect_loop, retry) is STOP_CONNECTING:
  File "/venv/lib/python2.7/site-packages/kazoo/retry.py", line 123, in __call__
    return func(*args, **kwargs)
  File "/venv/lib/python2.7/site-packages/kazoo/protocol/connection.py", line 512, in _connect_loop
    status = self._connect_attempt(host, port, retry)
  File "/venv/lib/python2.7/site-packages/kazoo/protocol/connection.py", line 539, in _connect_attempt
    read_timeout, connect_timeout = self._connect(host, port)
  File "/venv/lib/python2.7/site-packages/kazoo/protocol/connection.py", line 646, in _connect
    client._session_callback(KeeperState.CONNECTED)
  File "/venv/lib/python2.7/site-packages/kazoo/client.py", line 467, in _session_callback
    self._make_state_change(KazooState.CONNECTED)
  File "/venv/lib/python2.7/site-packages/kazoo/client.py", line 440, in _make_state_change
    remove = listener(state)
  File "/venv/lib/python2.7/site-packages/kazoo/recipe/cache.py", line 179, in _session_watcher
    self._root.on_reconnected()
  File "/venv/lib/python2.7/site-packages/kazoo/recipe/cache.py", line 217, in on_reconnected
    child.on_reconnected()
  File "/venv/lib/python2.7/site-packages/kazoo/recipe/cache.py", line 215, in on_reconnected
    self._refresh()
  File "/venv/lib/python2.7/site-packages/kazoo/recipe/cache.py", line 247, in _refresh
    self._refresh_data()
  File "/venv/lib/python2.7/site-packages/kazoo/recipe/cache.py", line 251, in _refresh_data
    self._call_client('get', self._path)
  File "/venv/lib/python2.7/site-packages/kazoo/recipe/cache.py", line 264, in _call_client
[...]

PR #683, opened 2022-11-12
feat(core): Introduce client rate limiting (concurrent_request_limit)

### LINKED ISSUE #664: TreeCache recipe creating heavy load on ZK while reconnecting.
The TreeCache recipe seems to generate a large amount of traffic while reconnecting to the ZK node ( This usually happens when there is a leader election in the cluster, which forces every client to get disconnected from the cluster and then go into a reconnection loop, which lasts until the leader election finishes. )

https://github.com/python-zk/kazoo/blob/9bb849941deb1d62ca3ca4d74882ed41cf569513/kazoo/recipe/cache.py#L210

Here we can see that the TreeCache tries to reload the entire Tree after getting reconnected to the ZK node. To test this, we ran a 3 node ZK cluster with 5k znodes and around 200 TreeCache clients (The clients use TreeCache to maintain the view of ZK data in memory ). During the leader election process, we noticed a 40 to 50x increase in read traffic (read traffic going from 600rps to 30k rps). This ends up causing a thundering herd problem for the cluster.

After some discussion, we came up with a few ideas that can probably help us to avoid the situation

1. Instead of reloading the entire Tree after a reconnect event, Only reload the znodes that have been updated between client disconnect and reconnect event. Zk guarantees that watches for the znodes will be triggered once clients get reconnected. we can use this to do a selective update of the znodes.
2. Introduce a client-side rate limiting which can help us to smooth the traffic burst after a reconnection. Clients should be able to use this to avoid overwhelming the cluster. 

Let me know what you think about the issue.

ceache:
Fixes #664
Relates to #458

## Why is this needed?

During connection failovers, cluster leader elections, or when clients manage large watch hierarchies (such as with `TreeCache`), clients can generate an unbounded surge of `get` and `get_children` requests.

ZooKeeper servers enforce an incoming request queue limit (`globalOutstandingLimit`, typically 1,000 requests across all connected clients). When multiple clients or heavy recipes reconnect simultaneously, this sudden flood can saturate server receive queues, trigger request drops or timeouts, and induce cascading reconnect loops across the ensemble (thundering herd). Similarly, event-driven applications reacting to high-frequency external triggers can easily overwhelm the server if in-flight requests are not bounded.

## Proposed Changes

- **Opt-in `concurrent_request_limit` Configuration**:
  - Added `concurrent_request_limit: PositiveInt | None = None` to `KazooClient.__init__` (default is `None`, which disables rate limiting and preserves 100% backward compatibility).
  - Strongly typed using `PositiveInt = Annotated[int, _Gt(0)]` under `if TYPE_CHECKING:`.
  - Added runtime validation raising `ConfigurationError("concurrent_request_limit must be greater than 0")` if configured with a non-positive value.

- **Wire-Level Throttling (Non-blocking & Deadlock-free)**:
  - Instead of blocking the calling thread or completion worker with a semaphore in `_call()` (which previously risked completion worker deadlocks on chained callbacks and double-release crashes on connection drops), rate limiting is implemented transparently at the transport layer in `ConnectionHandler._connect_attempt`.
  - The connection's `select()` loop checks `len(client._pending) < client.concurrent_request_limit`. When the in-flight wire limit is reached, `_read_sock` is excluded from `select()`.
  - Incoming responses from the ZooKeeper server are prioritized. The instant a response packet arrives and is popped from `_pending`, capacity frees up and the next queued request is sent over TCP.

- **Preserved Asynchronous Contract**:
  - All asynchronous calls (`get_async`, `create_async`, etc.) return immediately with an `IAsyncResult`.
  - Completion worker callbacks and watch handlers can freely issue further async requests without deadlocking.

- **Clean Drain on Connection Loss**:
  - In the event of a disconnect, both `_pending` and `_queue` are cleanly notified and drained with `ConnectionLoss` without leaking state or causing double-release errors.

## Related Issues & Discussions

- **#664**: `TreeCache recipe creating heavy load on ZK while reconnecting.` — Directly addresses the reconnect surge by smoothing traffic bursts and preventing ensemble overload.
- **#458**: `Deadlock with TreeCache and reconnection` — Background context on request burst handling during cache reconnection.

## Testing & Quality Assurance

- Added 7 dedicated unit tests in `kazoo/tests/test_rate_control.py`:
  1. Client initialization and argument validation (`None`, positive int, and `ConfigurationError` on non-positive values).
  2. Asynchronous calls never block the caller.
  3. Completion worker executing chained requests without deadlock.
  4. Connection loss drains both pending and queued requests cleanly.
  5. Select loop socket gating based on in-flight count.
  6. Pipeline flow simulation validating that in-flight requests never exceed the configured limit.
  7. `TreeCache` reconnect simulation (6 nodes, 12 concurrent operations) verifying deadlock-free completion under `concurrent_request_limit=1`.
- Clean static analysis:
  - `mypy -p kazoo` passes with 0 issues across 65 source files.
  - `flake8 kazoo` passes with 0 warnings/errors.

### DIFF
--- a/kazoo/client.py
+++ b/kazoo/client.py
@@ -120,6 +120,7 @@ class KazooClient(object):
         ca=None,
         use_ssl=False,
         verify_certs=True,
+        concurrent_request_limit=0,
         **kwargs,
     ):
         """Create a :class:`KazooClient` instance. All time arguments
@@ -241,6 +242,18 @@ class KazooClient(object):
         self.keyfile = keyfile
         self.keyfile_password = keyfile_password
         self.ca = ca
+        if concurrent_request_limit > 0:
+            self.logger.info(
+                "Zookeeper client rate-limited to %d concurrent requests",
+                concurrent_request_limit,
+            )
+            self.rate_limiting_sem = self.handler.semaphore_impl(
+                concurrent_request_limit
+            )
+
+        else:
+            self.rate_limiting_sem = None
+
         # Curator like simplified state tracking, and listeners for
         # state transitions
         self._state = KeeperState.CLOSED
@@ -634,6 +647,16 @@ class KazooClient(object):
         elif self._state == KeeperState.EXPIRED_SESSION:
             async_object.set_exception(SessionExpiredError())
             return False
+
+        if self.rate_limiting_sem:
+            if not self.rate_limiting_sem.acquire(blocking=False):
+                self.logger.info(
+                    "Limiting concurrent requests. Waiting for completion."
+                )
+                # Actually block on the sempahore here
+                self.rate_limiting_sem.acquire(blocking=True)
+            # Register the release of the semaphore on async request completion
+            async_object.rawlink(lambda _res: self.rate_limiting_sem.release())
 
         self._queue.append((request, async_object))
 
--- a/kazoo/handlers/eventlet.py
+++ b/kazoo/handlers/eventlet.py
@@ -11,6 +11,7 @@ from eventlet.green import time as green_time
 from eventlet.green import threading as green_threading
 from eventlet.green import selectors as green_selectors
 from eventlet import queue as green_queue
+from eventlet import semaphore as green_semaphore
 
 from kazoo.handlers import utils
 from kazoo.handlers.utils import selector_select
@@ -80,6 +81,7 @@ class SequentialEventletHandler(object):
     name = "sequential_eventlet_handler"
     queue_impl = green_queue.LightQueue
     queue_empty = green_queue.Empty
+    semaphore_impl = green_semaphore.BoundedSemaphore
 
     def __init__(self):
         """Create a :class:`SequentialEventletHandler` instance"""
--- a/kazoo/handlers/gevent.py
+++ b/kazoo/handlers/gevent.py
@@ -1,4 +1,5 @@
 """A gevent based handler."""
+
 from __future__ import absolute_import
 
 import atexit
@@ -14,7 +15,10 @@ import gevent.selectors
 
 from kazoo.handlers.utils import selector_select
 
-from gevent.lock import Semaphore, RLock
+from gevent.lock import (
+    BoundedSemaphore as Semaphore,
+    RLock as RLock,
+)
 
 from kazoo.handlers import utils
 
@@ -52,6 +56,7 @@ class SequentialGeventHandler(object):
     queue_impl = gevent.queue.Queue
     queue_empty = gevent.queue.Empty
     sleep_func = staticmethod(gevent.sleep)
+    semaphore_impl = Semaphore
 
     def __init__(self):
         """Create a :class:`SequentialGeventHandler` instance"""
--- a/kazoo/handlers/threading.py
+++ b/kazoo/handlers/threading.py
@@ -10,6 +10,7 @@ environments that use threads.
     :class:`~kazoo.handlers.gevent.SequentialGeventHandler` instead.
 
 """
+
 from __future__ import absolute_import
 
 import atexit
@@ -95,6 +96,7 @@ class SequentialThreadingHandler(object):
     sleep_func = staticmethod(time.sleep)
     queue_impl = queue.Queue
     queue_empty = queue.Empty
+    semaphore_impl = threading.BoundedSemaphore
 
     def __init__(self):
         """Create a :class:`SequentialThreadingHandler` instance"""


ceache:
@python-zk/maintainers This is a first draft. Let me know if you like the direction and I'll add tests, etc.

Full disclosure, we have had a version of this patch in production for a while (with both gevent and threading handlers). I have not tested the eventlet handler directly.

StephenSorriaux:
I think it is a great idea. Just wondering: from a user experience perspective, will there be any way to know the rate limit has been hit?

ceache:
No, at least not in this current form.

Sync requests will block, as if the zk server was slow.

This was designed for async requests so that code, possibly in different
threads, could simply queue them up knowing that the client would respect
the set rate limit. Think of chains of get_children_async | get_async when
walking a hierarchy.

We could add logging, it would not be difficult, but I would not want it to
be intrusive (i.e. debug?)

I feel like throttling for  rate limit is not an "issue", it is a feature.

Does thay make sense?

StephenSorriaux:
Yes, that makes a lot of sense, thank you.

I agree with you, rate limit is a feature, but sometimes it can be hard to "understand"/be aware of it if there is no logs telling you "hey, rate limit is currently being triggered" or "hey, rate limit has been triggered N times in the past M seconds" (even if, I agree, there is already a `info` log at client startup). You're right this should not be intrusive, I know some projects display a "rate limit has been triggered" message once in a while (every 30s or so for instance), but I believe a debug message can work too if you think it is better/useful.

Writing this message, it makes me realize that this lib currently does not provide any metrics (if it would, a number of times rate limit has been triggered would be useful). Maybe it is something we could add to the backlog, but that is not the point of your current proposition that, again, I think is great.

ceache:
I'll add a log message for now, and move to add some tests.

About metrics, i was talking to opentelemetry folks at KubeCon about
exactly that.
it looks like it would be possible to use such a framework. I mean, they
instrument SQLite3 and requests modules for example

IMHO, of we can pull this off, this would be an ideal way to extract this
kind of information, as well as request counts, error counts, connection
length and a million other things i have always wanted to know about my
zookeeper client but never knew how to ask 😜

On a more serious note, if you like the idea, i can create an issue for
this and try to gather more Intel?

kazoo/client.py
```diff
@@ -640,6 +653,10 @@ def _call(self, request, async_object):
             async_object.set_exception(SessionExpiredError())
             return False
 
+        if self.rate_limiting_sem:
+            self.rate_limiting_sem.acquire(blocking=True)
+            async_object.rawlink(lambda _res: self.rate_limiting_sem.release())
```
a-ungurianu, line 658: I saw we have some "recursive" async calls. Is there a way this can deadlock?

I think because we raw link to the Zookeeper response itself, the lambda being called, and then any other links that are used to implement these recursive calls are afterwards?

a-ungurianu: We can probably test this by parametrizing one of the async test suits to accept a KazooClient with the limit set to 1

ceache: We have in `kazoo.client` some chained async calls (`makepath=True` for example).
I am not sure what you mean by recursive otherwise.

The ZK protocol guarantees that request are always executed in order (no reordering), so I think this is safe no matter what, no?

kazoo/client.py
```diff
@@ -125,6 +125,7 @@ def __init__(
         ca=None,
         use_ssl=False,
         verify_certs=True,
+        max_async_requests=0,
```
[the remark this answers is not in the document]
ceache: You are correct.  I a not attached to the naming of the option at all... I would like it if it was less wordy but it needs to describe what it is. Completely open to suggestions

kazoo/client.py
a-ungurianu: The makepath was what I was thinking, yeah. Struggling to phrase it, but what I was thinking was that we might be sensitive to the order of the callbacks in https://github.com/python-zk/kazoo/blob/master/kazoo/handlers/utils.py#L133. 

Now that I look a bit further, the queue where we add the requests and the queue where the completions of the requests go are separate so I think it's ok.

StephenSorriaux:
> I'll add a log message for now, and move to add some tests. About metrics, i was talking to opentelemetry folks at KubeCon about exactly that. it looks like it would be possible to use such a framework. I mean, they instrument SQLite3 and requests modules for example IMHO, of we can pull this off, this would be an ideal way to extract this kind of information, as well as request counts, error counts, connection length and a million other things i have always wanted to know about my zookeeper client but never knew how to ask stuck_out_tongue_winking_eye On a more serious note, if you like the idea, i can create an issue for this and try to gather more Intel?

Sorry for my late reply @ceache, busy weeks :(

To be honest, I really like the idea and it would be great if you can get some intel about it, especially on how we should make those metrics available to our users: should we provide some interface so that they can plug whatever client they want? Should we actually provide some integration ourselves (like prometheus, etc.)? I totally feel you on this subject, I also love to get tons of metrics about what's in production!

kazoo/client.py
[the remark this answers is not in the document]
StephenSorriaux: FWIW, I really like Alex's suggestion.

Buffer0x7cd:
Hey Folks, any update on the progress of this PR ? Let me know if there is any help needed to complete this feature.

kazoo/client.py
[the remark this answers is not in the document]
ceache: done

ceache:
@python-zk/maintainers  I think I have implemented all what we discussed.  All more future looking ideas (prometheus, OTL,) would be addressed in future PRs

kazoo/client.py
```diff
@@ -635,6 +648,16 @@ def _call(self, request, async_object):
             async_object.set_exception(SessionExpiredError())
             return False
 
+        if self.rate_limiting_sem:
```
StephenSorriaux, line 651: can a test be added to check that we get blocked and then get released?

PR #688, opened 2022-12-28
Fix possible endless wait in stop() after AUTH_FAILED error

azat:
In case of AUTH_FAILED in the zk-loop thread it will call client._session_callback which will reset the queue.

However another thread can add to this queue CloseInstance event, and if the _session_callback() will be called after CloseInstance was added to the queue, then stop() will never return (and zk-loop will endlessly spin).

Here is how it looks like with addititional logging:

    39: [ Thread-3 (zk_loop) ] INFO: client.py:568: _session_callback: Zookeeper session closed, state: AUTH_FAILED
    39: [ MainThread ] Level 5: client.py:721: stop: Sending CloseInstance
    39: [ Thread-3 (zk_loop) ] Level 5: client.py:403: _reset: Reseting the client
    39: [ Thread-3 (zk_loop) ] Level 5: connection.py:625: _connect_attempt: Connecting
    39: [ Thread-3 (zk_loop) ] Level 5: connection.py:625: _connect_attempt: Connecting

You can find details in this gist [1].

  [1]: https://gist.github.com/azat/bc7aaea1c32a4f1ea75ad646d26280e9

### DIFF
--- a/kazoo/protocol/connection.py
+++ b/kazoo/protocol/connection.py
@@ -619,7 +619,7 @@ class ConnectionHandler(object):
             self.ping_outstanding.clear()
             last_send = time.time()
             with self._socket_error_handling():
-                while True:
+                while not self.client._stopped.is_set():
                     # Watch for something to read or send
                     jitter_time = random.randint(1, 40) / 100.0
                     deadline = last_send + read_timeout / 2.0 - jitter_time
--- a/kazoo/tests/test_client.py
+++ b/kazoo/tests/test_client.py
@@ -256,6 +256,8 @@ class TestAuthentication(KazooTestCase):
         with pytest.raises(AuthFailedError):
             client.add_auth("unknown-scheme", digest_auth)
 
+        client.stop()
+
     def test_add_auth_on_reconnect(self):
         client = self._get_client()
         client.start()


a-ungurianu:
Hey!

This makes sense in theory. Could you rebase in master so the tests rerun. Would appreciate a review from someone else, as I'm not super familiar with the lifetime cycle of the client

azat:
Rebased, but now it requires workflow approval

azat:
Looks like all failures are unrelated, since the last error in the log is:

```
==> Uploading
    .url https://codecov.io/
    .query commit=d9a20a766ba942b5722ee36a1210e62fbb0ae9f8&branch=HEAD&package=py2.1.12
    Gzipping contents..
    Compressed contents to 18443 bytes
    Pinging Codecov...
Error: Could not determine repo and owner
Tip: See all example repositories: https://github.com/codecov?query=example
.pkg: _exit> python /opt/hostedtoolcache/Python/3.10.9/x64/lib/python3.10/site-packages/pyproject_api/_backend.py True setuptools.build_meta
  py310-gevent-eventlet-sasl: FAIL code 1 (171.97=setup[19.90]+cmd[152.06] seconds)
  codecov: OK (14.40=setup[13.73]+cmd[0.66] seconds)
  evaluation failed :( (187.83 seconds)
Error: Process completed with exit code 255.
```

StephenSorriaux:
Yes you are right, we have some flaky tests...

azat:
So, any news/feedback on this one?

jeffwidman (approved):
Like Alex, this seems reasonable to me, but I'm not super familiar with this section of code.

I'm okay merging, but @ceache if you do get a chance I'd appreciate a sanity review on this since IIRC you run kazoo in prod scenarios that include auth... so may be more familiar with this.

jeffwidman:
@azat can you check the "allow edits from maintainers" box? I can't rebase to merge w/o that... or you can simply rebase it yourself, up to you.

jeffwidman (approved):

azat:
@jeffwidman I don't have `Allow edits from mainters` box due to the [github issues for pull requests from organizations](https://github.com/orgs/community/discussions/5634), and I completely forgot about this, next time will keep it in mind (maybe will do this from my main account).

Thanks!

jeffwidman:
Gotcha, I didn't even realize that was a functionality gap as I always contribute from my personal GitHub... makes sense.

PR #689, opened 2023-01-19
chore: add mypy to the build

bringhurst:
## Why is this needed?

This is intended to be a lightweight version of https://github.com/python-zk/kazoo/pull/684. Rather than add types to everything in one shot, support for typing could be added to the build only. Python code typing could be incrementally added in future smaller/targeted PRs.

The changes in https://github.com/python-zk/kazoo/pull/684 are still useful. This is just an effort to split it up to make it easier to understand.

The method used here follows the same pattern that SQLAlchemy has been using to add types. For an example, see https://github.com/sqlalchemy/sqlalchemy/pull/9039/files and note the removal of the `# mypy: ignore-errors` comment to incrementally add types.

## Proposed Changes

  - Add mypy configuration to the build to allow for incremental addition of type annotations.

No. This PR will only contain build-time type checking.

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -41,7 +41,7 @@ jobs:
       - name: Code check
         run: tox -e ${TOX_VENV}
         env:
-          TOX_VENV: black,pep8
+          TOX_VENV: black,pep8,mypy
 
   test:
     needs: [validate]
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -21,3 +21,125 @@ extend-exclude = '''
 
 [tool.pytest.ini_options]
 addopts = "-ra -v"
+
+[tool.mypy]
+
+# For details on each flag, please see the mypy documentation at:
+# https://mypy.readthedocs.io/en/stable/config_file.html#config-file
+
+# Note: The order of flags listed here should match the order used in mypy's
+# documentation to make it easier to find the documentation for each flag.
+
+# Import Discovery
+ignore_missing_imports = false
+
+# Disallow dynamic typing
+disallow_any_unimported = true
+disallow_any_expr = false
+disallow_any_decorated = true
+disallow_any_explicit = true
+disallow_any_generics = true
+disallow_subclassing_any = true
+
+# Untyped definitions and calls
+disallow_untyped_calls = true
+disallow_untyped_defs = true
+disallow_incomplete_defs = true
+check_untyped_defs = true
+disallow_untyped_decorators = true
+
+# None and Optional handling
+implicit_optional = false
+strict_optional = true
+
+# Configuring warnings
+warn_redundant_casts = true
+warn_unused_ignores = true
+warn_no_return = true
+warn_return_any = true
+warn_unreachable = true
+
+# Miscellaneous strictness flags
+allow_untyped_globals = false
+allow_redefinition = false
+local_partial_types = true
+implicit_reexport = false
+strict_concatenate = true
+strict_equality = true
+strict = true
+
+# Configuring error messages
+show_error_context = true
+show_column_numbers = true
+hide_error_codes = false
+pretty = true
+color_output = true
+error_summary = true
+show_absolute_path = true
+
+# Miscellaneous
+warn_unused_configs = true
+verbosity = 0
+
+# FIXME: As type annotations are introduced, please remove the appropriate
+#        ignore_errors flag below. New modules should NOT be added here!
+
+[[tool.mypy.overrides]]
+module = [
+    'kazoo.client',
+    'kazoo.exceptions',
+    'kazoo.handlers.eventlet',
+    'kazoo.handlers.gevent',
+    'kazoo.handlers.threading',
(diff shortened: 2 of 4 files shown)


bringhurst:
This appears to currently be failing an unrelated issue. See https://github.com/python-zk/kazoo/pull/691

bringhurst:
Hey @a-ungurianu -- please let me know if you have any thoughts on splitting up the type hinting PR like this. :)

a-ungurianu:
> Hey @a-ungurianu -- please let me know if you have any thoughts on splitting up the type hinting PR like this. :)

Haven't looked at the content of the PR, but conceptually I have no issues with it. Will review later today.

a-ungurianu:
Thanks a lot for the PR!

When it comes to introducing typings to a new repo, I've seen 2 orthogonal approaches, which is either rule by rule, or module by module.

In the past I've found success with the later approach, as it defines the target a bit more thoroughly and it also creates what I like to call a "ball of coherency", where we can (mostly) trust the interactions between the typed modules.

In theory I wouldn't be against this approach either, but having relatively lax rules might result in us publishing types that aren't accurate.

bringhurst:
That sounds good to me.

I was looking around at how some other projects are doing it -- it looks like SQLAlchemy is adding a type ignore to the top of each file, then slowly removing it as they go.

PR updated to use that method.

a-ungurianu:
If we want to go with the file-by-file approach, then we need to make the mypy requirements stricter.

I've pulled the broken bit out of #684 so it should be ready to review. This utilizes the configuration that mypy offers to configure which files are checked or not. I think I like the approach this PR offers for ignoring the files, as it is very visible when editing the file, whereas the mypy configuration is a bit farther away.

Would you be open to get the rule list from #684? After that we can merge this and then rebase #684 and also unlock the ability to parallelize the type checking between multiple contributors.

bringhurst:
> I've pulled the broken bit out of https://github.com/python-zk/kazoo/pull/684 so it should be ready to review. This utilizes the configuration that mypy offers to configure which files are checked or not. I think I like the approach this PR offers for ignoring the files, as it is very visible when editing the file, whereas the mypy configuration is a bit farther away.

I just switched it to use the centralized mypy config method. However, I ended up inverting the list of files to ignore so the config file will shrink in size over time (rather than grow to include all files in the repo).

> If we want to go with the file-by-file approach, then we need to make the mypy requirements stricter.

I just went through the mypy config line-by-line (based on the mypy docs) and made it as strict as possible (with the exception of `disallow_any_expr`, which looks like it's going to be unreasonably noisy until all external packages have types). If it turns out to be too strict, we can always relax it later. :)

Regarding ignores for external packages, for example:

```
module = 'gevent.*'
module = 'eventlet.*'
module = 'puresasl.*'
```

We should probably depend on  using `# type: ignore` comments on the import of modules with missing types -- along with the `warn_unused_ignores = true` flag set. That way, when the external package is upgraded to a version which includes typing, the build will fail with a decent error message until the `# type: ignore` comment is removed. If these external packages are ignored in the centralized config, types will silently be ignored when the external package is upgraded to a version which includes types (until someone notices somehow and removes it from the mypy.ini).

a-ungurianu:
Do you have a strong opinion pro having the separate `mypy.ini` file? I tend to not be a huge fan of the proliferation of dot-files, so given that we can now fold it into `pyproject.yaml` (see #684), I'd tend towards that

mypy.ini
```diff
@@ -0,0 +1,226 @@
+[mypy]
+
+# For details on each flag, please see the mypy documentation at:
+# https://mypy.readthedocs.io/en/stable/config_file.html#config-file
+
+# Note: The order of flags listed here should match the order used in mypy's
+# documentation to make it easier to find the documentation for each flag.
```
a-ungurianu, line 7: I like this. Thanks for adding the reference link. Always takes me 5 min or so to find the reference 😛

jeffwidman:
Thanks guys for working on this, I am not closely following the implementation details, but at a high-level I'm 👍 👍 !

bringhurst:
> Do you have a strong opinion pro having the separate mypy.ini file? I tend to not be a huge fan of the proliferation of dot-files, so given that we can now fold it into pyproject.yaml (see https://github.com/python-zk/kazoo/pull/684), I'd tend towards that

No strong preference at all -- I just updated the PR to have the config in the `pyproject.toml` instead (and also checked to make sure it still works). :)

a-ungurianu (approved):
LGTM. Thanks a lot and sorry for the back and forth :)

jeffwidman (approved):
This looks very reasonable to me, but I haven't been working professionally in Python the last few years, so haven't kept up with the idiomatic type tooling recommendations... so before we merge, I'd like to get @ofek's input, as I'm pretty sure he's stayed abreast of such things. I pinged him for a quick review, and assuming he's 👍 then we should merge.

jeffwidman:
Thanks again @bringhurst !

PR #691, opened 2023-01-19
chore: migrate to tox 4.x

bringhurst:
## Why is this needed?

tox-wheel is a deprecated package now that tox 4.x includes wheel support.

## Example error message

When this issue occurs, you may see an error message similar to:

```
  File "<frozen importlib._bootstrap>", line 1006, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 688, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 883, in exec_module
  File "<frozen importlib._bootstrap>", line 241, in _call_with_frames_removed
  File "/opt/hostedtoolcache/Python/3.10.9/x64/lib/python3.10/site-packages/tox_wheel/plugin.py", line 6, in <module>
    import py
ModuleNotFoundError: No module named 'py'
Error: Process completed with exit code 1.
```

## Proposed Changes

  - remove dependency on tox-wheel
  - use tox 4.x

No. Build time changes only.

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -36,7 +36,7 @@ jobs:
       - name: Install required dependencies
         run: |
           python3 -m pip install --upgrade pip
-          pip install tox tox-wheel
+          pip install tox
 
       - name: Code check
         run: tox -e ${TOX_VENV}
--- a/tox.ini
+++ b/tox.ini
@@ -1,8 +1,7 @@
 [tox]
-minversion = 3.24.1
+minversion = 4.3.4
 requires=
     virtualenv>=20.7.2
-    tox-wheel>=0.6.0
 skip_missing_interpreters=True
 envlist =
     pep8,black,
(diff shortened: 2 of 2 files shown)


bringhurst:
Note that it's likely this change also affects the release.yml github actions workflow in some way. I'm unsure on how to test that however.

StephenSorriaux:
Hello,

Thank you for this.

The `release.yml` workflow is only triggered from a tag, since it performs all the packaging and pypi upload stuff. Maybe we could add a `package.yml` workflow where we do the packaging stuff without performing the final pypi upload. In any cases, I feel like it should not be part of this PR. I can look into it, unless you prefer/have time to do it.

bringhurst:
After looking at it a bit more, I agree that it shouldn't be part of this PR.

I'll throw together another PR to get the conversation going on including packaging as part of CI (without publish).

StephenSorriaux (approved):
Thank you

jeffwidman (approved):

jeffwidman:
Going to merge as don't see a reason to hold off on this. Thanks @bringhurst !

PR #692, opened 2023-01-19
chore: add packaging without publish to CI

bringhurst:
## Why is this needed?

CI currently does not exercise packaging for pypi.

## Proposed Changes

  - Add a github actions workflow to test packaging (except for a pypi publish)

No.


bringhurst:
At the moment, this appears to be failing on unrelated issues. See https://github.com/python-zk/kazoo/pull/691

jeffwidman:
I merged #691 (thank you for that!), so this is ready to be rebased / finish out...

bringhurst:
The force push is to get rid of the merge commit.

StephenSorriaux (approved):
Thank you!

jeffwidman (changes requested):
So this looks fine, but it's also a heavy copy-paste of https://github.com/python-zk/kazoo/blob/1c0c4535c9c82092d94ce36fb7a0be4c43483bb4/.github/workflows/release.yml

The problem will be that if changes happen here or there, they may not be reflected to the other copy.

Could we instead chain the actions such that the steps are DRY'd up... one workflow does the build, the other workflow does the publish, and can trigger the build as part of it's initial steps?

I'm not an actions expert, but a quick google makes it look like this might be possible?
https://docs.github.com/en/actions/using-workflows/reusing-workflows

StephenSorriaux:
@jeffwidman Re-thinking about this, it seems we should actually publish "dev" version to the TestPyPI repository (using a version name like `X.Y.Z.dev+{short_commit_id}` or anything else if someone has a better idea) so that every step are tested every time. It would require a little more changes like adding a new secret for the repo.

Concerning Python 3.11 vs Python 3.10, I agree this is something that we can change once we add a pipeline for the 3.11 version of Python (we "only" have testing for up to 3.10 right now).

On my side, even if it is not perfect, I am fine going forward with the current state of the PR and doing the changes I mentioned above in another PR unless, of course, @bringhurst is interested in doing it on his side.

bringhurst:
I think it's probably ok to commit this as-is. It's not perfect, but it gets the job done.

> it seems we should actually publish "dev" version to the TestPyPI repository

I was looking at the `cookiecutter-hypermodern-python` project to to get ideas on how to do that, but I haven't had time to really dig in. They seem to have a very nice setup.

https://github.com/cjolowicz/cookiecutter-hypermodern-python/tree/main/{{cookiecutter.project_name}}/.github/workflows

jeffwidman:
Totally fine to punt on publishing to TestPyPI and `3.11`, but I'd _really_ rather we didn't introduce more copy/paste when we don't need to... that just increases risk of divergence in the build steps down the line resulting in fake confidence that our packaging is correct.

AFAICT, it should be straightforward to wire up actions such that we've got _one_ workflow that handles all builds, and then a second workflow takes the output from that and actually publishes it...

PR #693, opened 2023-01-21
Ask Dependabot to bump actions

jeffwidman:
I noticed some of our GitHub action workflows are throwing deprecation warnings...

Let's have Dependbot open PR's to bump them whenever they're outdated.

We could also enable it for watching our Python deps, but that may be more controversial, so that can be done as a follow-on PR if there's interest.


StephenSorriaux (approved):
Great idea! Thank you

PR #700, opened 2023-01-29
chore: add a per test 3min timeout and enable color output

bringhurst:
## Why is this needed?

It's difficult to reproduce, but I've seen the github actions test suite take more than 6 hours. Having a timeout set will allow us to get some sort of useful debug information instead of having github actions kill it off before we can analyze anything.

## Proposed Changes

 - Do not allow any one unit test to take more than an hour. If it takes longer than an hour, kill it off and print some basic debug info (e.g. a stack trace of all/most frames).
 - Enable color pytest output to make pytest easier to read in general.

No -- this is intended to help debug the test suite.

### DIFF
--- a/constraints.txt
+++ b/constraints.txt
@@ -5,3 +5,4 @@ flake8==5.0.2
 objgraph==3.5.0
 pytest==6.2.5
 pytest-cov==3.0.0
+pytest-timeout==2.2.0
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -20,11 +20,13 @@ extend-exclude = '''
 '''
 
 [tool.pytest.ini_options]
-addopts = "-ra -v"
+addopts = "-ra -v --color=yes"
 log_cli = true
 log_cli_date_format = "%Y-%m-%d %H:%M:%S"
 log_cli_format = "%(asctime)s %(levelname)s %(message)s"
 log_cli_level = "INFO"
+# Per-test timeout in seconds
+timeout = 180
 
 [tool.mypy]
 
--- a/setup.cfg
+++ b/setup.cfg
@@ -58,6 +58,7 @@ test =
     objgraph
     pytest
     pytest-cov
+    pytest-timeout
     gevent>=1.2 ; implementation_name!='pypy'
     eventlet>=0.17.1 ; implementation_name!='pypy'
     pyjks


bringhurst:
I keep seeing this error also (it appears to be a string returned from the codecov api):

```Could not determine repo and owner```

🤔

**Comment by ceache** [comment]
I think the timeout is ok.

If the tests did not complete in 1h, they are not going to complete ever.

**Comment by bringhurst** [comment]
Example run that seems like it'll probably hit the timeout:

https://github.com/python-zk/kazoo/actions/runs/4038631107/jobs/6942766386

**Comment by a-ungurianu** [comment]
> This organization is not using Codecov’s [GitHub App Integration](https://github.com/apps/codecov). We recommend you install it so Codecov can continue to function properly for your repositories. [Learn more](https://about.codecov.io/blog/codecov-is-updating-its-github-integration/?utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr+comments&utm_term=python-zk)

I wonder if this is why you see the codecov things, as it tries leverage the Github App.

@ceache I don't have the rights to add the app. Do you know someone who does?

**Comment by bringhurst** [comment]
An example of this showing useful debug output:

https://github.com/python-zk/kazoo/actions/runs/4038897905/jobs/6943223749

**Comment by bringhurst** [comment]
It looks like this job may time out -- https://github.com/python-zk/kazoo/actions/runs/4039044495/jobs/6943463553

If that works correctly (times out and shows stack traces), I'll open this up for review.

**Review by jeffwidman — DISMISSED** [review]
Happy to try it, and keep it as long as it's useful!

I do wonder whether this is just actions infra being flaky from time to time, or if we've got some bugs in our code that are hiding... could just be Zookeeper occasionally flaking as it comes up, I know I've seen that from time to time although no idea if that's Zookeeper, or Docker or something else...

File: tox.ini [hunk 1]
```diff
@@ -28,6 +28,7 @@ extras =
     eventlet: eventlet
     sasl: sasl
 deps =
+    pytest-timeout
```
a-ungurianu, line 31: I see both pytest and pytest-cov are in the `constraints.txt` file at the root level of the repo, so probably it'd be good to add this there as well for consistency.

StephenSorriaux:
I have resurrected this PR and reduced the `timeout` from 1h to 3min since it is a per-test timeout.

@jeffwidman @a-ungurianu @ceache if you have some time to check it

jeffwidman (approved):
Looks great 👍

ceache (approved):

PR #702, opened 2023-01-30
No need to specify `mock`

jeffwidman:
It's now part of the Python stdlib since `3.3`.

So no need to specify it.

### DIFF
--- a/constraints.txt
+++ b/constraints.txt
@@ -2,7 +2,6 @@
 black==22.10.0
 coverage==6.3.2
 flake8==5.0.2
-mock==3.0.5
 objgraph==3.5.0
 pytest==6.2.5
 pytest-cov==3.0.0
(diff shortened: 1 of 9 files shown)


a-ungurianu:
`mock` is indeed available in stdlib, but it's under [`unittest.mock`](https://docs.python.org/3/library/unittest.mock.html)

jeffwidman:
Thanks @a-ungurianu, I threw this PR up quickly from web UI and then completely forgot about it!

jeffwidman:
This is ready for review. The miniscule drop in coverage is a side effect of changing imports and shouldn't block merging.

jeffwidman:
Can I please get a review on this so we can land it before merge conflicts creep in?

StephenSorriaux (approved):
Thank you!

PR #704, opened 2023-01-31
debug: try bumping mock version

jeffwidman:
Seeing a weird error in https://github.com/python-zk/kazoo/pull/702/ and wondering if it's because latest `mock` is actually `5.0.1`: https://mock.readthedocs.io/en/latest/changelog.html

so threw this up to see what CI thinks

### DIFF
--- a/constraints.txt
+++ b/constraints.txt
@@ -2,7 +2,7 @@
 black==22.10.0
 coverage==6.3.2
 flake8==5.0.2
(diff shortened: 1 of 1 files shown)


jeffwidman:
Nope, the problem was in my grep'ing I only looked for `import mock` not `from mock import`... so missed some of the necessary import renames.

PR #705, opened 2023-01-31
refactor: delete backup travis config

jeffwidman:
We seem pretty happy with GitHub Actions, and now that Travis is owned by PE their product seems to have gone downhill so even if we were looking for an alternative, I don't think Travis would  be it. Regardless, this backup will still remain in git history, so no point in duplicating version control via `.bak` files.


a-ungurianu (approved):
Sounds good!

PR #715, opened 2023-03-21
feat: add support for persistent recursive watches

jeblair:
## Why is this needed?

ZooKeeper 3.6.0 added support for persistent, and persistent recursive watches.  This adds the corresponding support to the Kazoo client class.

No breaking changes.

### DIFF
--- a/kazoo/client.py
+++ b/kazoo/client.py
@@ -25,6 +25,7 @@ from kazoo.loggingsupport import BLATHER
 from kazoo.protocol.connection import ConnectionHandler
 from kazoo.protocol.paths import _prefix_root, normpath
 from kazoo.protocol.serialization import (
+    AddWatch,
     Auth,
     CheckVersion,
     CloseInstance,
@@ -38,6 +39,7 @@ from kazoo.protocol.serialization import (
     SetACL,
     GetData,
     Reconfig,
+    RemoveWatches,
     SetData,
     Sync,
     Transaction,
@@ -48,6 +50,8 @@ from kazoo.protocol.states import (
     KazooState,
     KeeperState,
     WatchedEvent,
+    AddWatchMode,
+    WatcherType,
 )
 from kazoo.retry import KazooRetry
 from kazoo.security import ACL, OPEN_ACL_UNSAFE
@@ -248,6 +252,8 @@ class KazooClient(object):
         self.state_listeners = set()
         self._child_watchers = defaultdict(set)
         self._data_watchers = defaultdict(set)
+        self._persistent_watchers = defaultdict(set)
+        self._persistent_recursive_watchers = defaultdict(set)
         self._reset()
         self.read_only = read_only
 
@@ -416,8 +422,16 @@ class KazooClient(object):
         for data_watchers in self._data_watchers.values():
             watchers.extend(data_watchers)
 
+        for persistent_watchers in self._persistent_watchers.values():
+            watchers.extend(persistent_watchers)
+
+        for pr_watchers in self._persistent_recursive_watchers.values():
+            watchers.extend(pr_watchers)
+
         self._child_watchers = defaultdict(set)
         self._data_watchers = defaultdict(set)
+        self._persistent_watchers = defaultdict(set)
+        self._persistent_recursive_watchers = defaultdict(set)
 
         ev = WatchedEvent(EventType.NONE, self._state, None)
         for watch in watchers:
@@ -1644,8 +1658,111 @@ class KazooClient(object):
 
         return async_result
 
+    def add_watch(self, path, watch, mode):
+        """Add a watch.
+
+        This method adds persistent watches.  Unlike the data and
+        child watches which may be set by calls to
+        :meth:`KazooClient.exists`, :meth:`KazooClient.get`, and
+        :meth:`KazooClient.get_children`, persistent watches are not
+        removed after being triggered.
+
+        To remove a persistent watch, use
+        :meth:`KazooClient.remove_all_watches` with an argument of
+        :attr:`~kazoo.protocol.states.WatcherType.ANY`.
+
+        The `mode` argument determines whether or not the watch is
+        recursive.  To set a persistent watch, use
+        :class:`~kazoo.protocol.states.AddWatchMode.PERSISTENT`.  To set a
+        persistent recursive watch, use
+        :class:`~kazoo.protocol.states.AddWatchMode.PERSISTENT_RECURSIVE`.
+
+        :param path: Path of node to watch.
+        :param watch: Watch callback to set for future changes
+                      to this path.
+        :param mode: The mode to use.
+        :type mode: int
+
+        :raises:
+            :exc:`~kazoo.exceptions.MarshallingError` if mode is
+            unknown.
+
+            :exc:`~kazoo.exceptions.ZookeeperError` if the server
+            returns a non-zero error code.
+        """
+        return self.add_watch_async(path, watch, mode).get()
+
+    def add_watch_async(self, path, watch, mode):
+        """Asynchronously add a watch. Takes the same arguments as
+        :meth:`add_watch`.
+        """
+        if not isinstance(path, str):
+            raise TypeError("Invalid type for 'path' (string expected)")
+        if not callable(watch):
+            raise TypeError("Invalid type for 'watch' (must be a callable)")
+        if not isinstance(mode, int):
+            raise TypeError("Invalid type for 'mode' (int expected)")
+        if mode not in (
+            AddWatchMode.PERSISTENT,
+            AddWatchMode.PERSISTENT_RECURSIVE,
+        ):
+            raise ValueError("Invalid value for 'mode'")
+
+        async_result = self.handler.async_result()
+        self._call(
+            AddWatch(_prefix_root(self.chroot, path), watch, mode),
+            async_result,
+        )
+        return async_result
+
+    def remove_all_watches(self, path, watcher_type):
+        """Remove watches from a path.
+
+        This removes all watches of a specified type (data, child,
+        any) from a given path.
+
+        The `watcher_type` argument specifies which type to use.  It
+        may be one of:
+
+        * :attr:`~kazoo.protocol.states.WatcherType.DATA`
+        * :attr:`~kazoo.protocol.states.WatcherType.CHILDREN`
+        * :attr:`~kazoo.protocol.states.WatcherType.ANY`
+
+        To remove persistent watches, specify a watcher type of
+        :attr:`~kazoo.protocol.states.WatcherType.ANY`.
+
+        :param path: Path of watch to remove.
+        :param watcher_type: The type of watch to remove.
+        :type watcher_type: int
+        """
+
+        return self.remove_all_watches_async(path, watcher_type).get()
+
+    def remove_all_watches_async(self, path, watcher_type):
+        """Asynchronously remove watches. Takes the same arguments as
+        :meth:`remove_all_watches`.
+        """
+        if not isinstance(path, str):
+            raise TypeError("Invalid type for 'path' (string expected)")
+        if not isinstance(watcher_type, int):
+            raise TypeError("Invalid type for 'watcher_type' (int expected)")
+        if watcher_type not in (
+            WatcherType.ANY,
+            WatcherType.CHILDREN,
+            WatcherType.DATA,
+        ):
+            raise ValueError("Invalid value for 'watcher_type'")
+
+        async_result = self.handler.async_result()
+        self._call(
+            RemoveWatches(_prefix_root(self.chroot, path), watcher_type),
+            async_result,
+        )
+        return async_result
+
 
 class TransactionRequest(object):
+
     """A Zookeeper Transaction Request
 
     A Transaction provides a builder object that can be used to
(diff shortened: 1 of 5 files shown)


jeblair:
It looks like the test failures are not related to this change.

kazoo/protocol/connection.py
```diff
@@ -363,6 +365,18 @@ def _write(self, msg, timeout):
                     raise ConnectionDropped("socket connection broken")
                 sent += bytes_sent
 
+    def _find_persistent_watchers(self, path):
```
westphahl, line 368: Do we have to distinguish between persistent and persistent recursive watches here? Given a persistent (non-recursive) watch for `/a` and another one for `/a/b` it looks like an event for `/a/b` would currently also trigger the callback for the watch for `/a`.

kt-315:
Hi! Why not merge this?

westphahl (approved):

kt315 (approved):

kt-315:
Hi! Whose approval is need? 🤔

StephenSorriaux:
Hello,

Approval from @python-zk/maintainers is needed. Will try to take a look by end of the week.

Can somebody that already tried the feature can confirm it is working as expected?

jeblair:
> Hello,
> 
> Approval from @python-zk/maintainers is needed. Will try to take a look by end of the week.
> 
> Can somebody that already tried the feature can confirm it is working as expected?

While awaiting review/merge we in the Zuul project have vendored a copy[1] of this and used it successfully in production for about nine months (though it does look like our vendored copy omits the ability to remove watches since we never do so, so we'll have to rely on the tests for that).  Thanks for taking a look.  :)

[1] https://opendev.org/zuul/nodepool/src/branch/master/nodepool/zk/vendor

kazoo/client.py
```diff
@@ -1644,8 +1656,100 @@ def reconfig_async(self, joining, leaving, new_members, from_config):
...
+
+        This method adds persistent watches.  Unlike the data and
+        child watches which may be set by calls to
+        :meth:`KazooClient.exists`, :meth:`KazooClient.get`, and
+        :meth:`KazooClient.get_children`, persistent watches are not
+        removed after being triggered.
+
+        To remove a persistent watch, use
+        :meth:`KazooClient.remove_all_watches` with an argument of
+        :attr:`~kazoo.states.WatcherType.ANY`.
```
StephenSorriaux, line 1670: i think all references to `~kazoo.states` should be replaced with `~kazoo.protocol.states`

kazoo/tests/test_client.py
```diff
@@ -1160,6 +1165,118 @@ def test_request_queuing_session_expired(self):
...
+        client.set("/a/b/d", value=b"2")
+        client.delete("/a", recursive=True)
+        # Remove the watch
+        client.remove_all_watches("/a", WatcherType.ANY)
+        # Perform one more call that we don't expect to see
+        client.create("/a")
+        # Wait in case the callback does arrive
+        time.sleep(3)
+        assert events == [
+            dict(type="CREATED", path="/a"),
```
StephenSorriaux, line 1205: should we use `kazoo.protocol.states.EventType.CREATED` instead (and `CHANGED/DELETED` for the ones below?

StephenSorriaux (commented):
Thank you for the PR, the implementation is great, I just added some questions. Would it be possible to take into account the warning from codecov and cover some of the new lines with a test?

kazoo/tests/test_client.py
```diff
@@ -1160,6 +1165,118 @@ def test_request_queuing_session_expired(self):
...
+        def data_callback(event):
+            pass
+
+        client = self.client
+        client.create("/a")
+        client.get("/a", watch=data_callback)
+        client.get_children("/a", watch=child_callback)
+        client.remove_all_watches("/a", WatcherType.DATA)
+        client.create("/a/b")
+        callback_event.wait(30)
```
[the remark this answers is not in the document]
StephenSorriaux: answering to myself: no, it does not make sense as the timeout should not be triggered

kazoo/tests/test_client.py
```diff
@@ -1160,6 +1165,118 @@ def test_request_queuing_session_expired(self):
...
+        def child_callback(event):
+            pass
+
+        client = self.client
+        client.create("/a")
+        client.get("/a", watch=data_callback)
+        client.get_children("/a", watch=child_callback)
+        client.remove_all_watches("/a", WatcherType.CHILDREN)
+        client.set("/a", b"1")
+        callback_event.wait(30)
```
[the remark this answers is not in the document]
StephenSorriaux: answering to myself: no, it does not make sense as the timeout should not be triggered

StephenSorriaux:
@jeffwidman @ceache @a-ungurianu this one is now ready to be reviewed
@jeblair I made some small changes, mostly based on my previous review, please let me now if you see anything wrong

StephenSorriaux (dismissed):

jeblair:
> @jeffwidman @ceache @a-ungurianu this one is now ready to be reviewed @jeblair I made some small changes, mostly based on my previous review, please let me now if you see anything wrong

The changes LGTM, thanks!  And sorry I wasn't able to address those myself in a timely manner.

StephenSorriaux:
Thank you @jeblair 

@jeffwidman @ceache @a-ungurianu gentle bump!

kazoo/client.py
```diff
@@ -1644,8 +1658,111 @@ def reconfig_async(self, joining, leaving, new_members, from_config):
...
+        async_result = self.handler.async_result()
+        self._call(
+            RemoveWatches(_prefix_root(self.chroot, path), watcher_type),
+            async_result,
+        )
+        return async_result
+
 
 class TransactionRequest(object):
+
```
jeffwidman, line 1765: Is this blank line extraneous? I don't normally recall a blank line between class definition and docstring...

kazoo/protocol/connection.py
```diff
@@ -448,13 +468,35 @@ def _read_response(self, header, buffer, offset):
...
+                if watcher:
+                    if isinstance(request, AddWatch):
+                        if request.mode == AddWatchMode.PERSISTENT:
+                            client._persistent_watchers[request.path].add(
+                                watcher
+                            )
+                        elif request.mode == AddWatchMode.PERSISTENT_RECURSIVE:
+                            client._persistent_recursive_watchers[
+                                request.path
+                            ].add(watcher)
```
jeffwidman, line 483: I wonder if this might benefit from an `else` clause?

I realize it's a no-op, but it still took me a moment to realize that's because it might be a non-persistant watch that's being added... so might be easier for a reader to understand with a code comment... 

But code comments can go out of sync, so a simple assertion that it's the expected mode might be useful like:
```python
else:
    if request.mode not in [expected modes]:
       raise <relevant exception>
```

that way both clarifies the code and also catches stupid errors if somehow a case was missed down the road.

Alternatively, this might read even simpler as a switch statement with a default case... but I realize those require Python 3.10 so maybe leave a:
```python
# TODO: switch these if/else clauses to switch statements where appropriate once we drop Python 3.9 support
```

kazoo/protocol/connection.py
```diff
@@ -448,13 +468,35 @@ def _read_response(self, header, buffer, offset):
...
+                    elif isinstance(request, (GetChildren, GetChildren2)):
+                        client._child_watchers[request.path].add(watcher)
+                    else:
+                        client._data_watchers[request.path].add(watcher)
+                if isinstance(request, RemoveWatches):
+                    if request.watcher_type == WatcherType.CHILDREN:
+                        client._child_watchers.pop(request.path, None)
+                    elif request.watcher_type == WatcherType.DATA:
+                        client._data_watchers.pop(request.path, None)
+                    elif request.watcher_type == WatcherType.ANY:
```
jeffwidman, line 493: ~this might read simpler as a switch statement rather than the `if/elif/elif` series? technically it should be more performant too~

Oh, NVM, just realized switch statements won't work in Python 3.8/3.9 😢

kazoo/protocol/states.py
```diff
@@ -251,3 +251,44 @@ def data_length(self):
...
     def children_count(self):
         return self.numChildren
+
+
+class AddWatchMode(object):
+    """Modes for use with :meth:`~kazoo.client.KazooClient.add_watch`
+
+    .. attribute:: PERSISTENT
+
+        The watch is not removed when trigged.
```
jeffwidman, line 261:
```suggestion
        The watch is not removed when triggered.
```

kazoo/protocol/states.py
```diff
@@ -251,3 +251,44 @@ def data_length(self):
...
+
+    .. attribute:: PERSISTENT_RECURSIVE
+
+        The watch is not removed when trigged, and applies to all
```
jeffwidman, line 265:
```suggestion
        The watch is not removed when triggered, and applies to all
```

kazoo/tests/test_client.py
```diff
@@ -1158,6 +1164,147 @@ def test_request_queuing_session_expired(self):
         finally:
             client.stop()
 
+    def _require_zk_version(self, major, minor):
```
jeffwidman, line 1167: This would benefit from a docstring describing the goal of the function... since the `pytest.skip` marker isn't until the end of the method, stating at the top "this function skips the test if the zookeeper version isn't greater than or equal to major.minor" would allow a dev to skim it faster.

kazoo/tests/test_client.py
```diff
@@ -1158,6 +1164,147 @@ def test_request_queuing_session_expired(self):
...
+            skip = False
+        else:
+            ver = self.client.server_version()
+            if ver[1] < minor:
+                skip = True
+        if skip:
+            pytest.skip("Must use Zookeeper %s.%s or above" % (major, minor))
+
+    def test_persistent_recursive_watch(self):
+        # This tests adding and removing a persistent recursive watch.
```
jeffwidman, line 1181:
```suggestion
        """This tests adding and removing a persistent recursive watch."""
````

File: kazoo/tests/test_client.py [hunk 13]
```diff
@@ -1158,6 +1164,147 @@ def test_request_queuing_session_expired(self):
...
+            dict(type=EventType.CHANGED, path="/a/b/c"),
+            dict(type=EventType.CHANGED, path="/a/b/d"),
+            dict(type=EventType.DELETED, path="/a/b/c"),
+            dict(type=EventType.DELETED, path="/a/b/d"),
+            dict(type=EventType.DELETED, path="/a/b"),
+            dict(type=EventType.DELETED, path="/a"),
+        ]
+
+    def test_persistent_watch(self):
+        # This tests adding and removing a persistent watch.
```
jeffwidman, line 1220:
```suggestion
        """This tests adding and removing a persistent watch."""
````

File: kazoo/tests/test_client.py [hunk 14]
```diff
@@ -1158,6 +1164,147 @@ def test_request_queuing_session_expired(self):
...
+            dict(type=EventType.CREATED, path="/a/b/d"),
+            dict(type=EventType.CHANGED, path="/a/b/c"),
+            dict(type=EventType.CHANGED, path="/a/b/d"),
+            dict(type=EventType.DELETED, path="/a/b/c"),
+            dict(type=EventType.DELETED, path="/a/b/d"),
+            dict(type=EventType.DELETED, path="/a/b"),
+            dict(type=EventType.DELETED, path="/a"),
+        ]
+
+    def test_persistent_watch(self):
```
jeffwidman, line 1219: nit: I'd probably order the tests:
1. `test_persistent_watch`
2. `test_persistent_recursive_watch`

As 2️⃣ feels like a variant of 1️⃣ , so typically we put the simpler one first.

kazoo/tests/test_client.py
```diff
@@ -1158,6 +1164,147 @@ def test_request_queuing_session_expired(self):
...
+        # Wait in case the callback does arrive
+        time.sleep(3)
+        assert client._persistent_watchers[full_path] == set()
+        assert events == [
+            dict(type=EventType.CREATED, path="/a"),
+            dict(type=EventType.DELETED, path="/a"),
+        ]
+
+    def test_remove_data_watch(self):
+        # Test that removing a data watch leaves a child watch in place.
```
jeffwidman, line 1248:
```suggestion
        """Test that removing a data watch leaves a child watch in place."""
````

File: kazoo/tests/test_client.py [hunk 16]
```diff
@@ -1158,6 +1164,147 @@ def test_request_queuing_session_expired(self):
...
+        client = self.client
+        client.create("/a")
+        client.get("/a", watch=data_callback)
+        client.get_children("/a", watch=child_callback)
+        client.remove_all_watches("/a", WatcherType.DATA)
+        client.create("/a/b")
+        callback_event.wait(30)
+
+    def test_remove_children_watch(self):
+        # Test that removing a children watch leaves a data watch in place.
```
jeffwidman, line 1267:
```suggestion
        """Test that removing a children watch leaves a data watch in place."""
````

jeffwidman (dismissed):
Looks good!

I made a few mostly cosmetic suggestions, but functionally it all looks good to me.

There's enough complexity in the protocol that even with the nice unit test coverage it wouldn't surprise me if there was a bug or two in here, either on our end or zookeeper's end, but given the solid UT coverage I think the best way to flush them out at this point is to merge and let folks bang on it for a while. The nice part is that this is a new feature, so any problems are unlikely to affect existing functionality.

jeffwidman:
Yes, please update the PR.

PR #716, opened 2023-03-24
Make tests less flaky on CI

StephenSorriaux:
## Why is this needed?
This makes the tests way less flaky, I hope (and first results confirmed it... but you never know).

## Proposed Changes
  - renamed back `test__connection.py` to `test_connection.py`, where it belongs
  - added `client_cluster_health` in the `setup_zookeeper` step that is there to "guarantee" the cluster is up and running, without even considering the configuration that will be used by the test itself. I added a "custom" retry logic without using a `KazooRetry` since `KazooRetry` does not handle refused/dropped connections like I would want too: even if a `connection_retry` is given to `KazooClient()`, the retry configuration is actually not respected because the `interrupt()` end up being triggered anyway (ie. the `client._stopped` event)
  - now performing a ZK cluster nuking at the end of each test case (not each test, it's different) since keeping the same ZK state from test cases to test cases actually led to some tests randomly failing
  - fixed the `test_read_only` test based on what is done in the official java client (https://github.com/apache/zookeeper/blob/master/zookeeper-server/src/test/java/org/apache/zookeeper/test/ReadOnlyModeTest.java), we were missing some things and that led the test to fail from time to time (see comments in the code)
  - activated the pytest `log_cli` configuration, as it gives more logs for each tests and I find it useful

Nope.


jeffwidman:
I ❤️ the effort, and the changes look mostly good, but there's a _lot_ of disparate changes in here... would it be possible at a minimum to separate some of them out to distinct commits, at ideally to separate PR's?

Benefits:
* easier to revert only the broken bits w/o reverting everything
* easier to identify broken bits via `git bisect` since more granular
* easier to understand for reviews now
* easier for `git blame` spelunkers later to understand the context of individual lines w/o needing to understand a massive PR

A few obvious ones that should be easy to peel off:
* the `coverage` changes, great catch btw! 👍 
* the `test_read_only` changes

In fact TBH, most of those bullet points look like obvious candidates for distinct PRs.

I'm happy to 🙋 volunteer to review the separate PR's and try to do so quickly so you're not blocked / slow to get them merged.

StephenSorriaux:
Thanks Jeff for taking the time. I will split my stuff.

kazoo/tests/conftest.py
```diff
@@ -4,8 +4,10 @@
...
 
 def pytest_exception_interact(node, call, report):
-    if hasattr(node._testcase, "cluster"):
+    try:
         cluster = node._testcase.cluster
         log.error("Zookeeper cluster logs:")
         for logs in cluster.get_logs():
             log.error(logs)
+    except Exception:
+        log.exception("Cant get ZK logs:")
```
jeffwidman, line 13:
```suggestion
        log.exception("Cannot get ZK logs:")
```

kazoo/tests/test_connection.py
```diff
@@ -1,4 +1,5 @@
 from collections import namedtuple, deque
+
```
jeffwidman, line 2: nit: why the newline?

pyproject.toml
```diff
@@ -21,6 +21,11 @@ extend-exclude = '''
 
 [tool.pytest.ini_options]
 addopts = "-ra -v"
+# XXX: uncomment if in need off all the captured logs with timing and stuff
```
jeffwidman, line 24:
```suggestion
# Note: uncomment to see the  logs with timing and data
```

However, this isn't actually commented out? So the comment doesn't match reality?

Should this instead default to `log_cli = false` and can be flipped to true? Rather than commenting/uncommenting?

kazoo/tests/test_connection.py
```diff
@@ -258,7 +258,7 @@ def back(state):
 class TestReadOnlyMode(KazooTestCase):
     def setUp(self):
         os.environ["ZOOKEEPER_LOCAL_SESSION_RO"] = "true"
-        self.setup_zookeeper(read_only=True)
+        self.setup_zookeeper()
```
jeffwidman, line 261: is `read_only=True` the default?

kazoo/tests/conftest.py
StephenSorriaux: Thanks, changed

kazoo/tests/test_connection.py
StephenSorriaux: No, this method takes some kwargs that will be passed to the kazoo client used in the test. This is not useful anymore, hence the removal

kazoo/testing/harness.py
```diff
@@ -208,8 +212,29 @@ def setup_zookeeper(self, **client_options):
...
+                tries += 1
+                if tries >= MAX_INIT_TRIES:
+                    raise
+                if tries > 0 and tries % 2 == 0:
+                    self.log(
+                        logging.WARNING,
+                        "nuking current cluster and making another one",
+                    )
+                    self.cluster.terminate()
+                    self.cluster.start()
```
jeffwidman, line 235: Is this nuking the underlying ZK cluster, or just the kazoo client trying to connect to the cluster? If the former, that's fairly expensive/slow, so do we need to sleep/pause a bit for it to get going before continuing? Or is that already handled internally within `cluster.start()`? (probably a better place for it)

kazoo/testing/harness.py
```diff
@@ -259,3 +284,9 @@ def setUp(self):
 
     def tearDown(self):
         self.teardown_zookeeper()
+
+    @classmethod
+    def tearDownClass(cls):
+        cluster = get_global_cluster()
```
jeffwidman, line 290: Oof, I have been bitten too many times by globals, I haven't looked at this, I suppose it probably makes sense here in testing, but def prefer in most cases passing around the actual instance to the class itself wherever possible.

kazoo/tests/conftest.py
```diff
@@ -4,8 +4,10 @@
 
 
 def pytest_exception_interact(node, call, report):
-    if hasattr(node._testcase, "cluster"):
```
jeffwidman, line 7: Why does this disappear? Because now we're always trying to get ZK logs... are there tests that don't have the "cluster" attr? If so, we're needlessly trying/failing to fetch the ZK logs IIUC...

kazoo/tests/test_connection.py
```diff
@@ -279,7 +278,15 @@ def test_read_only(self):
         from kazoo.exceptions import NotReadOnlyCallError
         from kazoo.protocol.states import KeeperState
 
-        client = self.client
+        if CI:
+            # force some wait to make sure the data produced during the
+            # `setUp()` step are replicaed to all zk members
```
jeffwidman, line 283: the casing of `setUp()` looks like Java, should it be changed to `setup()` ?? Or is that intentional... ?

kazoo/tests/test_connection.py
```diff
@@ -303,6 +311,11 @@ def listen(state):
                 thread.start()
             for thread in zk_stop_threads:
                 thread.join()
+            # stopping the client is *mandatory*, else the client might try to
+            # reconnect using a xid that the server may endlessly refuse
+            # restarting the client makes sure the xid is resetted
```
jeffwidman, line 316:
```suggestion
            # restarting the client makes sure the xid gets reset
```

kazoo/tests/test_lock.py
```diff
@@ -469,6 +469,10 @@ def test_write_lock(self):
         assert gotten is False
 
     def test_rw_lock(self):
+        pytest.skip(
+            "This test is flaky "
+            "(and actually performs some ZK behavior test...)"
+        )
```
jeffwidman, line 475: Is there benefit to keeping this test vs outright deleting it?

Always makes me scared to just drop a locking test, but if we move it to pending, it'll be out of sight, out of mind and likely never get fixed up... I recognize investigating the source of flakiness might be outside the scope of this PR and you already spent some time on it, but I still am tempted to say we either delete it or continue to let it flake intermittently rather than skipping... 🤷‍♂️

kazoo/tests/test_connection.py
jeffwidman, line 1: Did you research why this was originally `__`??

Was it a typo from the beginning, or an intentional choice to avoid `pytest` finding it automatically? Since `test_*` is a magic prefix for `pytest`.

jeffwidman (commented):
A few more comments. Overall, I'm still 👍 on this, but realized there's a few more sections where it'd be helpful to have more of the history/rationale of the change documented, not only for me but also future `git blame` spelunkers.

Would also appreciate if @ceache takes a look.

jeffwidman:
> * I added a "custom" retry logic without using a `KazooRetry` since `KazooRetry` does not handle refused/dropped connections like I would want too: even if a `connection_retry` is given to `KazooClient()`, the retry configuration is actually not respected because the `interrupt()` end up being triggered anyway (ie. the `client._stopped` event)

This is fine for this PR, but is this something we should consider changing/fixing in `KazooClient`/`KazooRetry` long-term?

kazoo/testing/harness.py
StephenSorriaux: Yes this global thing was there since forever...

`tearDownClass` is part of the `TestCase` class lifecycle, from the stdlib: https://docs.python.org/3/library/unittest.html#unittest.TestCase.tearDownClass, this helps performing the cluster cleanup between each test case

kazoo/tests/test_connection.py
StephenSorriaux: It is intentional as I am referring to the `setUp()` method from the `TestCase` class (https://docs.python.org/3/library/unittest.html#unittest.TestCase.setUp)

kazoo/tests/test_connection.py
StephenSorriaux: Yes, that was me, as part of https://github.com/python-zk/kazoo/pull/659, I noticed the test was less flaky when performed first, so I added a `_` to get it done first

kazoo/testing/harness.py
StephenSorriaux: Yes, there is already some `sleep()` hidden in `cluster.start()` (https://github.com/python-zk/kazoo/blob/master/kazoo/testing/common.py#L370)

kazoo/tests/conftest.py
StephenSorriaux: Without this change I was not able to get the ZK logs, for some reasons.

Examples:
- test that failed recently without this change, no "Zookeeper cluster logs": https://github.com/python-zk/kazoo/actions/runs/4661094003/jobs/8249964080
- test that failed with this change, you get some "Zookeeper cluster logs": https://github.com/python-zk/kazoo/actions/runs/4586252822/jobs/8098978225

Maybe I am missing something obvious

ceache:
I'll try to review this thoroughly this week. Sorry I have been quite
distracted recently.

Besides the flakiness within the tests themselves, i was looking at the
harness recently and I wanted to know how you felt about swapping our
custom zookeeper ensemble orchestration driver for some docker compose
driver.
This would remove a lot of complexity and allow us to know we are starting
the tests only after the ensemble is actually up and ready.
I know how to craft such a compose definition and used it numerous times
for my own testing.
The down side is introducing a dependency on docker for local testing (or
we could leave the "native" harness around for docker-less development)

What do you think?

kazoo/tests/test_lock.py
StephenSorriaux: Re-reading the test, I was wrong, it is actually a valid one. It is not skipped anymore.

kazoo/tests/test_connection.py
StephenSorriaux: Thanks

kazoo/tests/test_lock.py
StephenSorriaux: FWIW, the test was introduced in https://github.com/python-zk/kazoo/pull/650

StephenSorriaux:
> > * I added a "custom" retry logic without using a `KazooRetry` since `KazooRetry` does not handle refused/dropped connections like I would want too: even if a `connection_retry` is given to `KazooClient()`, the retry configuration is actually not respected because the `interrupt()` end up being triggered anyway (ie. the `client._stopped` event)
> 
> This is fine for this PR, but is this something we should consider changing/fixing in `KazooClient`/`KazooRetry` long-term?

Yes, I think we should consider looking into it, especially what to expect of the retry when used as a `connection_retry` parameter for the client

StephenSorriaux:
I'll be all in for a docker based harness! I personally have java installed on my desktop just for kazoo...

kazoo/testing/harness.py
jeffwidman: Oh that's right, I forgot we're still using `unittest` and not completely migrated over to `pytest`...

kazoo/tests/conftest.py
jeffwidman: What you've changed it to sounds reasonable to me... I was hoping the `"cluster"` attr would be auto-set, but looks like it's probably missing in some cases, and we don't want to have to remember to always manually set it.

jeffwidman:
I would 110% support a docker-based harness. No problem adding that as a test/dev dependency, it's almost certainly more accessible to drive-by devs as well who are going to be familiar with that from other libraries doing similar things.

jeffwidman (approved):
This all looks good to me, thank you so much for the detailed explanations, not only for me but also on behalf of other folks who stumble across this later.

Probably best to wait to merge til Charles get a chance to look at it too.

ceache (approved):
This looks good to me. Thanks a lot @StephenSorriaux !

StephenSorriaux:
> > > * I added a "custom" retry logic without using a `KazooRetry` since `KazooRetry` does not handle refused/dropped connections like I would want too: even if a `connection_retry` is given to `KazooClient()`, the retry configuration is actually not respected because the `interrupt()` end up being triggered anyway (ie. the `client._stopped` event)
> > 
> > 
> > This is fine for this PR, but is this something we should consider changing/fixing in `KazooClient`/`KazooRetry` long-term?
> 
> Yes, I think we should consider looking into it, especially what to expect of the retry when used as a `connection_retry` parameter for the client

Update: I think https://github.com/python-zk/kazoo/pull/685 tried to tackle the "issue" I mentionned

ceache:
I will try to rebase my PR, now that this is merged.

Would there be anything holding it back ?
I understood Jeff's comment to mean he thought it was ok.

StephenSorriaux:
Sorry for the rebase, `test_connection.py` might give you some conflict...

I think I owe you a review on https://github.com/python-zk/kazoo/pull/685, will take the time now!

PR #717, opened 2023-03-29
fix command just recv 8K

yyfyfyang:
Fixes #

## Why is this needed?
sock.recv(8192)
recv byte 8K, Can not be modified，there will be data loss

### DIFF
--- a/kazoo/client.py
+++ b/kazoo/client.py
@@ -769,7 +769,10 @@ class KazooClient(object):
             verify_certs=self.verify_certs,
         )
         sock.sendall(cmd)
-        result = sock.recv(8192)
+        result = b""
+        temp = sock.recv(8192)
+        while temp:
+            result += temp
         sock.close()
         return result.decode("utf-8", "replace")
 


jeffwidman:
Hmm, the description of the problem makes sense, I haven't dug into lower level networking code in a while, so not sure if this is the best way to solve or if this should be fixed through an OS-level config, or if 8k is python's default regardless of the OS setting... 

Regardless, now that #716 landed, can you rebase this to see what CI thinks?

kazoo/client.py
```diff
@@ -769,7 +769,10 @@ def command(self, cmd=b"ruok"):
             verify_certs=self.verify_certs,
         )
         sock.sendall(cmd)
-        result = sock.recv(8192)
+        result = b""
+        temp = sock.recv(8192)
+        while temp:
+            result += temp
```
ceache, line 775: This code loops forever as you never update `temp` in the loop

ceache (changes requested):

PR #719, opened 2023-04-10
feat(ci): remove `codecov` step from `tox` since performed in GA

StephenSorriaux:
## Why is this needed?
There is no need to perform the non-working `codecov` step via `tox` since we now have a Github Actions step that performs it.
## Proposed Changes

  - update `tox.ini` and remove `codecov` related stuff
  - update GA `testing.yml` to remove `codecov` from matrix

Nah.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -29,7 +29,6 @@ extras =
     sasl: sasl
 deps =
     sasl: kerberos
-    codecov: codecov
 allowlist_externals =
     {toxinidir}/ensure-zookeeper-env.sh
     {toxinidir}/init_krb5.sh
@@ -39,9 +38,6 @@ commands =
         pytest {posargs: -ra -v --cov-report=xml --cov=kazoo kazoo/tests}
 
 [testenv:build]
-
-[testenv:codecov]
-commands = - codecov -e TOX_VENV,ZOOKEEPER_VERSION
 
 [testenv:pep8]
 basepython = python3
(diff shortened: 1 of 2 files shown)


tox.ini
```diff
@@ -40,9 +39,6 @@ commands =
 
 [testenv:build]
 
-[testenv:codecov]
-commands = - codecov -e TOX_VENV,ZOOKEEPER_VERSION
```
jeffwidman, line 44: It looks like previously we ran coverage separately for each matrix entry... so removing this we'll only run it once, which means we lose visibility if a particular python/ zookeper version combo loses coverage unexpectedly due to a special casing in the code...

I personally think that's fine, coverage is more a "nice-to-have" tool to surface things, and rather have it run quickly than exhaustively, but wanted to document this.

jeffwidman (dismissed):
👍 

To run coverage locally, someone doesn't need to specify anything beyond `coverage`, right? No extra env vars / flags? Or do they need to specify `coverage -e ZOOKEPER_VERSION`?

StephenSorriaux:
> +1
> 
> To run coverage locally, someone doesn't need to specify anything beyond `coverage`, right? No extra env vars / flags? Or do they need to specify `coverage -e ZOOKEPER_VERSION`?

Locally, if you want to use `coverage` you would just need to perform `coverage xml`, yes.
FWIW `coverage` is "integrated" with `pytest`, in the `tox.ini` configuration (the `pytest-cov` dep + the `--cov` and `--cov-report` flags).

jeffwidman (approved):

PR #725, opened 2023-07-21
fix(log): add logger for kazoo.handlers.threading

### LINKED ISSUE #724: No handlers could be found for logger "kazoo.handlers.threading"
No handlers could be found for logger "kazoo.handlers.threading"

after using logging.basicConfig(), it turns out 
```
[2023-07-13 11:27:22,005] [ERROR] [3722:139833607571200] [threading.py:L130] Exception in worker queue thread
  File "/usr/lib/python2.7/site-packages/kazoo/handlers/threading.py", line 128, in _thread_worker
    func()
  File "/usr/lib/python2.7/site-packages/kazoo/handlers/threading.py", line 291, in <lambda>
    self.callback_queue.put(lambda: callback.func(*callback.args))
  File "/usr/local/services/myproject/src/zk.py", line 154, in leader_watch
    log.info("leader event: %s" % event)
TypeError: not all arguments converted during string formatting
```

but maybe it is better to has its own logger, just like kazoo.client does

zqfan:
Fixes #724

No handlers could be found for logger "kazoo.handlers.threading"

for better logging control, just like kazoo.client does

### DIFF
--- a/kazoo/client.py
+++ b/kazoo/client.py
@@ -221,7 +221,9 @@ class KazooClient(object):
         self.logger = logger or log
 
         # Record the handler strategy used
-        self.handler = handler or SequentialThreadingHandler(logger=self.logger)
+        self.handler = handler
+        if not handler:
+            self.handler = SequentialThreadingHandler(logger=self.logger)
         if inspect.isclass(self.handler):
             raise ConfigurationError(
                 "Handler must be an instance of a class, "
--- a/kazoo/client.py
+++ b/kazoo/client.py
@@ -221,7 +221,7 @@ class KazooClient(object):
         self.logger = logger or log
 
         # Record the handler strategy used
-        self.handler = handler if handler else SequentialThreadingHandler()
+        self.handler = handler or SequentialThreadingHandler(logger=self.logger)
         if inspect.isclass(self.handler):
             raise ConfigurationError(
                 "Handler must be an instance of a class, "
--- a/kazoo/handlers/threading.py
+++ b/kazoo/handlers/threading.py
@@ -110,6 +110,8 @@ class SequentialThreadingHandler(object):
         return self._running
 
     def _create_thread_worker(self, work_queue):
+        log = self.logger
+
         def _thread_worker():  # pragma: nocover
             while True:
                 try:
@@ -119,7 +121,7 @@ class SequentialThreadingHandler(object):
                             break
                         func()
                     except Exception:
-                        self.logger.exception("Exception in worker queue thread")
+                        log.exception("Exception in worker queue thread")
                     finally:
                         work_queue.task_done()
                         del func  # release before possible idle
--- a/kazoo/handlers/threading.py
+++ b/kazoo/handlers/threading.py
@@ -96,8 +96,9 @@ class SequentialThreadingHandler(object):
     queue_impl = queue.Queue
     queue_empty = queue.Empty
 
-    def __init__(self):
+    def __init__(self, logger=None):
         """Create a :class:`SequentialThreadingHandler` instance"""
+        self.logger = logger or log
         self.callback_queue = self.queue_impl()
         self.completion_queue = self.queue_impl()
         self._running = False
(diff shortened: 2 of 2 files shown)


ceache:
Hi @zqfan. Thank you for the PR but, I am sorry but I am not clear on the goals of this PR.
What is gained by making the handler reuse the client's logger?
Do you have a clear example?

zqfan:
> Hi @zqfan. Thank you for the PR but, I am sorry but I am not clear on the goals of this PR. What is gained by making the handler reuse the client's logger? Do you have a clear example?

https://github.com/python-zk/kazoo/blob/b00d88ff1485f1cb175565c49955bf5d7ce96fef/kazoo/handlers/threading.py#L121  This line needs log to be set, i.e.,  logging.basicConfig(), otherwise, it might print an error message on stderr: "No handlers could be found for logger "kazoo.handlers.threading"" . As issue https://github.com/python-zk/kazoo/issues/724 pointed out,  it leads to some information unnoticed, and cause a lot of time to dig out.

But it might be better if wen can specify a logger handler, which allows us to print log info into a specific file. Upper level application can only deal with kazoo.client, and it is not a good idea to expose kazoo/handlers/threading.py detail to upper level just for a logger handler, so I let it inherit the logger setting in kazoo.client.

this pr keeps same behavior as before, if user don't specify a dedicated logger handler

StephenSorriaux:
Thank you for the PR.

Just like @ceache said, I am also not sure about the goal of this PR.

I understand the issue about the `No handlers could be found for {logger}` (which is explained in the [kazoo documentation](https://kazoo.readthedocs.io/en/latest/basic_usage.html#logging-setup)) that, I think, can be solved using a default `NullHandler()` for our top logger (as per the [official documentation](https://docs.python.org/3/howto/logging.html#configuring-logging-for-a-library)).

However, I don't think passing a logger should be the way to go.

zqfan:
> Thank you for the PR.
> 
> Just like @ceache said, I am also not sure about the goal of this PR.
> 
> I understand the issue about the `No handlers could be found for {logger}` (which is explained in the [kazoo documentation](https://kazoo.readthedocs.io/en/latest/basic_usage.html#logging-setup)) that, I think, can be solved using a default `NullHandler()` for our top logger (as per the [official documentation](https://docs.python.org/3/howto/logging.html#configuring-logging-for-a-library)).
> 
> However, I don't think passing a logger should be the way to go.

My project doesn't configure a logging.basicConfig for everything, it adjust and control major module's behaviour separately. For now, to avoid too much message printed to my log file, I have to do something like this:

```
logger = logging.getLogger("kazoo.client")
logger.setLevel(logging.ERROR)  // for other module, it is INFO
for h in mymodule.logger.handlers:  // a different file for other module
    logger.addHandler(h)
self.client = KazooClient(hosts=ZK_HOSTS,logger=logger)
```

If kazoo.handlers.threading could not inherit logger settings from kazoo client, then I have to do it twice. 
I think end user should treat kazoo as a integrated module, settings should be done only once.

StephenSorriaux:
> If kazoo.handlers.threading could not inherit logger settings from kazoo client, then I have to do it twice.
I think end user should treat kazoo as a integrated module, settings should be done only once.

That can already be done: `kazoo.handlers.threading` does not inherit from `kazoo.client` but both inherit from `kazoo`, ie.

```python
import logging
logging.basicConfig(level=logging.INFO)
...
logger = logging.getLogger("kazoo")
logger.setLevel(logging.ERROR)
```

@ceache I think the possible use case for this, which is not explained here or in the initial issue, is the same as the reason why we have a `logger` configuration in the client (see https://github.com/python-zk/kazoo/pull/85): passing a `logger` around to possibly discriminate between multiple instances of `kazoo` running on the same host or whatever. I must say I am... intrigued: isn't that something the developer must take care during the development (via the formatter, etc.)? `boto3` or `requests`, to take some "famous" libs, do not let developers pass around their logger and I can easily understand why since Python already provides everything for that... Anyway, I would love your thoughts on this.

jeffwidman:
I understand what you're trying to achieve, but as @StephenSorriaux points out above, I think it's already solvable if you change the logger you're tuning to `kazoo` rather than `kazoo.client`.

May I recommend @brandon-rhodes excellent [`logging_tree` library](https://pypi.org/project/logging_tree/)? It's been a massive time saver for me over the years when I'm debugging unfamiliar code.

Closing, but if for some reason you can't achieve this with existing tooling please comment and we can re-evaluate.

PR #737, opened 2024-01-23
test: update Python and ZK in testing suite

### LINKED ISSUE #708: Meta: Minimum supported python version
I feel that it'd be useful to "codify" which python versions we want to support for this library.

Looking at https://devguide.python.org/versions/, python3.7 will drop off support in July this year, and python3.11 has been officially released for a few months.

The only other python implementation we test and support is pypy which [seems to have already removed pypy3.7 from their download page](https://www.pypy.org/download.html#!).

My suggesting is to provide a EoL + some buffer support after which we can safely remove it from the checks and be able to use the new features for the library (and playing our role in prodding library users to move off of unsupported and unsafe python version)

StephenSorriaux:
Fixes #708 

## Why is this needed?
[Python 3.7 is EOL since last year, Python 3.11 and Python 3.12 are here now.](https://devguide.python.org/versions/)
[ZK 3.6 is EOL since end of 2022 and ZK 3.9 is here.](https://zookeeper.apache.org/releases.html)

## Proposed Changes
  - drop Python 3.7 from testing
  - add Python 3.11 and 3.12 to testing
  - drop ZK 3.4 and 3.5
  - add ZK 3.8 and 3.9

No.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -13,10 +13,10 @@ jobs:
     - name: Handle the code
       uses: actions/checkout@v4
 
-    - name: Set up Python 3.10
+    - name: Set up Python 3.12
       uses: actions/setup-python@v5
       with:
-        python-version: "3.10"
+        python-version: "3.12"
 
     - name: Install pypa/build
       run: >-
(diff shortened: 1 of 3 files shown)


jeffwidman (dismissed):

jeffwidman:
Thanks 🙏

jeffwidman (approved):
👍 for merging w/o worrying about the decrease in coverage, it's a red herring here IMO.

Issue #727, opened 2023-08-08
RuntimeError: Invalid error code


nkormakov:
## Expected Behavior
No error, connection should be closed.

## Actual Behavior
RuntimeError: Invalid error code

## Snippet to Reproduce the Problem

```
zk = KazooClient(hosts = f'{ip}', verify_certs = False, timeout = 10)
zk.start()
zk.get_children(target_path)
```
It is happening when zookeeper running with a few extra jvm parameters
```
-Djava.security.auth.login.config=/etc/kafka/zookeeper_jaas.conf
-Dzookeeper.allowSaslFailedClients=false
-Dzookeeper.sessionRequireClientSASLAuth=true
```

Default library doesn't handle well error code -124 which declares that client should authorise himself. Our application is ok with closing connection altogether and not messing with it.
I've fixed it with monkey patching:
```
def new_read_response(self, header, buffer, offset):
    client = self.client
    request, async_object, xid = client._pending.popleft()
    if header.zxid and header.zxid > 0:
        client.last_zxid = header.zxid
    if header.xid != xid:
        exc = RuntimeError('xids do not match, expected %r '
                           'received %r', xid, header.xid)
        async_object.set_exception(exc)
        raise exc

    # Determine if it's an exists request and a no node error
    exists_error = (header.err == NoNodeError.code and
                    request.type == Exists.type)

    # Set the exception if it's not an exists error
    if header.err and not exists_error:
```
```
        # -124 error handler
        if header.err == -124:
            async_object.set(False)
            async_object.set_exception(NoAuthError())
            return CLOSE_RESPONSE
```
```
        callback_exception = EXCEPTIONS[header.err]()
        self.logger.debug(
            'Received error(xid=%s) %r', xid, callback_exception)
        if async_object:
            async_object.set_exception(callback_exception)
    elif request and async_object:
        if exists_error:
            # It's a NoNodeError, which is fine for an exists
            # request
            async_object.set(None)
        else:
            try:
                response = request.deserialize(buffer, offset)
            except Exception as exc:
                self.logger.exception(
                    "Exception raised during deserialization "
                    "of request: %s", request)
                async_object.set_exception(exc)
                return
            self.logger.debug(
                'Received response(xid=%s): %r', xid, response)

            # We special case a Transaction as we have to unchroot things
            if request.type == Transaction.type:
                response = Transaction.unchroot(client, response)

            async_object.set(response)

        # Determine if watchers should be registered
[...]

PR #743, opened 2024-02-06
fix(core): Add missing Zookeeper exceptions codes

ceache:
Sync'd up to Zookeeper 3.9.2 error codes.

## Why is this needed?

Server errors with missing error codes are handled poorly and result in connection close.

## Proposed Changes

  - Add definition for all know error codes as of Zookeeper 3.9.2

No.

### DIFF
--- a/kazoo/exceptions.py
+++ b/kazoo/exceptions.py
@@ -1,4 +1,5 @@
 """Kazoo Exceptions"""
+
 from collections import defaultdict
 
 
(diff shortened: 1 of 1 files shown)


StephenSorriaux:
Nice! Does this relate to #727?

StephenSorriaux (approved):
Thank you!

ceache:
@StephenSorriaux  yes! It does resolve #727

Issue #582, opened 2020-01-15
Hangs on stop()


efimkin:
Hi!
Sometimes our application hangs on stop()
kazoo (2.6.1)
```
(gdb) py-bt
Traceback (most recent call first):
  <built-in method acquire of _thread.lock object at remote 0x7fac03095df0>
  File "/usr/lib/python3.6/threading.py", line 1072, in _wait_for_tstate_lock
    elif lock.acquire(block, timeout):
  File "/usr/lib/python3.6/threading.py", line 1056, in join
    self._wait_for_tstate_lock()
  File "/opt/yandex/pgsync/lib/python3.6/site-packages/kazoo/protocol/connection.py", line 193, in stop
    self._connection_routine.join()
  File "/opt/yandex/pgsync/lib/python3.6/site-packages/kazoo/client.py", line 528, in _safe_close
    if not self._connection.stop(timeout):
  File "/opt/yandex/pgsync/lib/python3.6/site-packages/kazoo/client.py", line 635, in stop
    self._safe_close()
  File "/opt/yandex/pgsync/lib/python3.6/site-packages/pgsync/zk.py", line 165, in __del__
    self._zk.stop()
```

azat:
I've also came across this issue, in my case the problem was in race after AUTH_FAILED error, here is a fix - #688

And here is a gist with details - https://gist.github.com/azat/bc7aaea1c32a4f1ea75ad646d26280e9

PR #744, opened 2024-02-22
Revert "Fix possible endless wait in stop() after AUTH_FAILED error (…

jeblair:
Revert "Fix possible endless wait in stop() after AUTH_FAILED error (#688)"
    
This reverts commit 5225b3e2fab6fec3b12b789e3cc6f3218429d32d.

The commit being reverted here caused kazoo not to empty the send
queue before disconnecting.  This means that if a client submitted
asynchronous requests and then called client.stop(), the connection
would be closed immediately, usually after only one (but possibly
more) of the submitted requests were sent.  Prior to this, Kazoo
would empty the queue of submitted requests all the way up to and
including the Close request when client.stop() was called.

Another area where this caused problems is in a busy multi-threaded
system.  One thread might decide to gracefully close the connection,
but if there is any traffic generated by another thread, then the
connection would end up terminating without ever sending the Close
request.

Failure to gracefully shutdown a ZooKeeper connection can mean that
other system components need to wait for ephemeral node timeouts to
detect that a component has shutdown.

Given that this behavior is easily reproducible and can have serious
consequences in production, 5225b3e2 is reverted.

### DIFF
--- a/kazoo/protocol/connection.py
+++ b/kazoo/protocol/connection.py
@@ -619,7 +619,7 @@ class ConnectionHandler(object):
             self.ping_outstanding.clear()
             last_send = time.monotonic()
             with self._socket_error_handling():
-                while not self.client._stopped.is_set():
+                while True:
                     # Watch for something to read or send
                     jitter_time = random.randint(1, 40) / 100.0
                     deadline = last_send + read_timeout / 2.0 - jitter_time
--- a/kazoo/tests/test_client.py
+++ b/kazoo/tests/test_client.py
@@ -258,8 +258,6 @@ class TestAuthentication(KazooTestCase):
         with pytest.raises(AuthFailedError):
             client.add_auth("unknown-scheme", digest_auth)
 
-        client.stop()
-
     def test_add_auth_on_reconnect(self):
         client = self._get_client()
         client.start()


jeblair:
Here is a test script that will show the behavior in both async and
threaded environments:

```
import kazoo.client
import time
import threading
import logging

logging.basicConfig(level=5)
l = logging.getLogger('kazoo')
l.setLevel(5)

client = kazoo.client.KazooClient('localhost:2181')
client.start()

def send():
    client.create('/test', b'',
                  makepath=True, ephemeral=True, sequence=True)

def test_thread():
    t = threading.Thread(target=send)
    t.start()
    client.stop()

def test_async():
    for x in range(100):
        client.create_async('/test', b'',
                            makepath=True, ephemeral=True, sequence=True)
    client.stop()

# test_thread()
test_async()
```

jeblair:
In addition to the reasoning in the commit message, I have also
attempted to reproduce the problem described in #688 in an attempt to
avoid regressing on that bug fix.  I have been unable to reproduce the
error using the script provided in that PR.  Since the PR description
mentioned threading, and the provided script did not appear to use it,
I also ran a test script I devised myself along with an added sleep
call in Kazoo to try to reproduce the behavior and was still unable to
do so.  In every case, it appears that the remote ZooKeeper server
closed the connection after the AUTH_FAILED response which caused the
Kazoo loop to terminate.  The script and changes I made can be seen at:
https://github.com/jeblair/kazoo/commit/7e22fea13fe4fd8b38fb34e5476ce4a32119a874

I tested with ZooKeeper 3.8.3; perhaps that is a behavior change from
when the bug was originally reported.

westphahl (approved):

StephenSorriaux:
Thank you for the PR, I will look into it.

(FWIW, tests failures seem unrelated to the change)

ceache (approved):
Looking at the original PR, it seems the change was not addressing the issue correctly. And quite honestly, I don't understand how the test case provided in the gist triggered the issue that was described.
The proper and expected behavior is that everything ahead of the close in the send queue be flushed before closing. So the reversal looks proper to me.

In any case, LGTM. Thanks!

StephenSorriaux (approved):
Thank you again for the PR.

You mentioned this might be an issue because of some specific ZK version, I was not able to reproduce it using ZK 3.6.4 or ZK 3.5.10.
Also, `stop()` and `close()` are called after every test methods (during `tearDown()`), so I think we would have detected it before if the issue was there, unfortunately the job logs has expired so I can't check that.

FWIW, the initial PR mentioned the issue https://github.com/python-zk/kazoo/issues/582 which more likely seem would be fixed by adding a `timeout` in the thread `join()`.

Anyway, I agree with both you and also think this should be reversed. Would it be possible to make sure the commit title match our guidelines?

ceache:
Thanks!

PR #747, opened 2024-03-03
fix(test): avoid racy reader vs writer contender in `test_rw_lock`

StephenSorriaux:
See test failure https://github.com/python-zk/kazoo/actions/runs/8116444130/job/22206401448?pr=744

```python
        for contender in ("reader", "writer"):
            thread, event = contender_bits[contender]
    
            with self.condition:
                while not self.active_thread:
                    self.condition.wait()
>               assert self.active_thread == contender
E               AssertionError: assert 'writer' == 'reader'
E                 - reader
E                 + writer

kazoo/tests/test_lock.py:515: AssertionError
```

## Why is this needed?
`test_rw_lock` is creating 2 threads (`reader` and `writer`) and, after being started, it is expected that `reader` is a contender before `writer`. In some busy systems (like the CI... it's always the CI!) this may not be true and lead to the test failure because `writer` can be a contender before `reader`. This commit makes sure that `reader` is always a contender before `writer`.

## Proposed Changes
  - wait for `reader` being a contender before starting `writer`
  - rewrote some minimal part of the test, just because I was there

No!

### DIFF
--- a/kazoo/tests/test_lock.py
+++ b/kazoo/tests/test_lock.py
@@ -487,27 +487,29 @@ class KazooLockTests(KazooTestCase):
         lock = self.client.WriteLock(self.lockpath, "test")
         lock.acquire()
 
+        wait = self.make_wait()
         reader_thread.start()
+        # make sure reader_thread is a contender before writer_thread
+        wait(lambda: len(lock.contenders()) == 2)
         writer_thread.start()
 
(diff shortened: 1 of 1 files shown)


jeffwidman (approved):
Makes sense to me. Nice cleanup.

ceache (approved):
Looks good to me!

PR #748, opened 2024-03-11
fix(core): Proper retry count in KazooRetry

ceache:
Make sure the number of attempts matches the `max_retry` parameter. Add unit tests to that effect.

## Why is this needed?

So that users can precisely control the number of retries

## Proposed Changes

  - Move the location of the attempt increase to before the check against `max_try` so we don't end up calling `max_tries + 1` times the function.

No, it shouldn't.

### DIFF
--- a/kazoo/retry.py
+++ b/kazoo/retry.py
@@ -133,10 +133,10 @@ class KazooRetry(object):
             except ConnectionClosedError:
                 raise
             except self.retry_exceptions:
+                self._attempts += 1
                 # Note: max_tries == -1 means infinite tries.
                 if self._attempts == self.max_tries:
                     raise RetryFailedError("Too many retry attempts")
-                self._attempts += 1
                 jitter = random.uniform(
                     1.0 - self.max_jitter, 1.0 + self.max_jitter
                 )
--- a/kazoo/tests/test_retry.py
+++ b/kazoo/tests/test_retry.py
@@ -1,89 +1,103 @@
-import unittest
+from unittest import mock
 
 import pytest
 
+from kazoo import exceptions as ke
+from kazoo import retry as kr
 
-class TestRetrySleeper(unittest.TestCase):
-    def _pass(self):
+
+def _make_retry(*args, **kwargs):
+    """Return a KazooRetry instance with a dummy sleep function."""
+
+    def _sleep_func(_time):
         pass
 
-    def _fail(self, times=1):
-        from kazoo.retry import ForceRetryError
-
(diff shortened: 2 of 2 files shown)


kazoo/tests/test_retry.py
```diff
@@ -1,89 +1,107 @@
-import unittest
-
 import pytest
 
 
-class TestRetrySleeper(unittest.TestCase):
```
StephenSorriaux, line 6: Can I ask why not using a class anymore? It seems to make the `_try_counts` handling a little more "simple" (ie. without the use of `global` + doing its reinit in the `setup()` method) but I might be missing something.

kazoo/retry.py
```diff
@@ -133,10 +133,10 @@ def __call__(self, func, *args, **kwargs):
             except ConnectionClosedError:
                 raise
             except self.retry_exceptions:
+                self._attempts += 1
                 # Note: max_tries == -1 means infinite tries.
                 if self._attempts == self.max_tries:
                     raise RetryFailedError("Too many retry attempts")
-                self._attempts += 1
```
StephenSorriaux, line 139: Great catch!

kazoo/retry.py
```diff
@@ -133,10 +133,10 @@ def __call__(self, func, *args, **kwargs):
             except ConnectionClosedError:
                 raise
             except self.retry_exceptions:
+                self._attempts += 1
                 # Note: max_tries == -1 means infinite tries.
```
StephenSorriaux, line 137: Do you see a way to test this case?

kazoo/tests/test_retry.py
ceache: I cleaned this up.
I removed the class to get rid of the unittest model.
I think I cleaned this up enough now that it is actually better.

kazoo/retry.py
ceache: I am not sure what you mean: what 'case'?
Do you mean that the exception is in the `self.exceptions` set?

StephenSorriaux: I was thinking of the not `Note: max_tries == -1 means infinite tries.`

StephenSorriaux (approved):

kazoo/tests/test_retry.py
jeffwidman: FWIW, I'm a fan of getting rid of the Unittest subclassing model whenever possibel... I personally find it clunky compared to the pytest syntax.

jeffwidman (approved):

PR #750, opened 2024-04-13
chore: Default ReadTheDocs to Ubuntu / Python `latest`

jeffwidman:
I noticed this because `3.7` is EOL'd by upstream, plus `24.04` will be released later this month (although may take RTD a bit longer to pull it in).

Looking at [the docs](https://docs.readthedocs.io/en/stable/config-file/v2.html), there's a `latest` variant for both.

It comes with a warning that using latest may unexpectedly break builds if the project isn't compatible... But I think for our use case the odds that it will break are extremely low (and in fact we'd want to know it's broken so we could fix it!) Plus if it starts breaking, it's not hard to temp pin it to an older version.

So all-in-all, both lower maintenance and faster realization of breakage if we default to latest on both of these.


jeffwidman:
Hmm... something weird with working on codepsaces... I'll try re-opening.

PR #755, opened 2024-06-27
Added support for SNI

### LINKED ISSUE #754: Add support for SNI hostnames
## What would you like to be added

Send the SNI header with TLS connections.

## Why is this needed

Currently when connecting to TLS hosts, only the IP is used. Some sites are behind passthrough proxies that proxy based upon hostname in the SNI header. Without support for SNI, these zookeeper ensembles cannot be reached via the library.

jessesightler-redhat:
Fixes #754

## Why is this needed?

Supports SNI

## Proposed Changes

Add server_hostname to the wrap call

No

### DIFF
--- a/kazoo/client.py
+++ b/kazoo/client.py
@@ -771,7 +771,7 @@ class KazooClient(object):
             keyfile=self.keyfile,
             keyfile_password=self.keyfile_password,
             verify_certs=self.verify_certs,
-            check_hostname=self.check_hostname
+            check_hostname=self.check_hostname,
         )
         sock.sendall(cmd)
         result = sock.recv(8192)
--- a/kazoo/client.py
+++ b/kazoo/client.py
@@ -183,6 +183,8 @@ class KazooClient(object):
         :param use_ssl: argument to control whether SSL is used or not
         :param verify_certs: when using SSL, argument to bypass
             certs verification
+        :param check_hostname: when using SSL, check the hostname
+            against the hostname in the cert
 
         Basic Example:
 
--- a/kazoo/client.py
+++ b/kazoo/client.py
@@ -120,6 +120,7 @@ class KazooClient(object):
         ca=None,
         use_ssl=False,
         verify_certs=True,
+        check_hostname=False,
         **kwargs,
     ):
         """Create a :class:`KazooClient` instance. All time arguments
@@ -237,6 +238,7 @@ class KazooClient(object):
 
         self.use_ssl = use_ssl
         self.verify_certs = verify_certs
+        self.check_hostname = check_hostname
         self.certfile = certfile
         self.keyfile = keyfile
         self.keyfile_password = keyfile_password
@@ -758,8 +760,10 @@ class KazooClient(object):
             raise ConnectionLoss("No connection to server")
 
         peer = self._connection._socket.getpeername()[:2]
+        peer_host = self._connection._socket.getpeername()[1]
         sock = self.handler.create_connection(
             peer,
+            hostname=peer_host,
             timeout=self._session_timeout / 1000.0,
             use_ssl=self.use_ssl,
             ca=self.ca,
@@ -767,6 +771,7 @@ class KazooClient(object):
             keyfile=self.keyfile,
             keyfile_password=self.keyfile_password,
             verify_certs=self.verify_certs,
+            check_hostname=self.check_hostname
         )
         sock.sendall(cmd)
         result = sock.recv(8192)
--- a/kazoo/handlers/utils.py
+++ b/kazoo/handlers/utils.py
@@ -241,8 +241,7 @@ def create_tcp_connection(
             context.load_default_certs(ssl.Purpose.SERVER_AUTH)
             if check_hostname and not verify_certs:
                 raise ValueError(
-                    "Error, if check_hostname is specified "
-                    + "verify_certs must be False"
+                    "verify_certs must be False when check_hostname is specified"
                 )
             # We must set check_hostname to False prior to setting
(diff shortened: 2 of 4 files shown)


jessesightler-redhat:
Formatting issues have been resolved. I'm not sure why one of the pypi tests failed with a segfault when all of the others passed.

kazoo/client.py
```diff
@@ -120,6 +120,7 @@ def __init__(
         ca=None,
         use_ssl=False,
         verify_certs=True,
+        check_hostname=False,
```
StephenSorriaux, line 123: Can you please add a description of this new param L185? This is useful for the documentation.

kazoo/handlers/utils.py
```diff
@@ -241,7 +243,7 @@ def create_tcp_connection(
             # verify_mode to CERT_NONE.
             # TODO: Make hostname verification configurable as some users may
```
StephenSorriaux, line 244: Can we remove this TODO since your changes make it completed?

kazoo/tests/test_utils.py
```diff
@@ -30,6 +30,24 @@ def test_timeout_arg(self):
                     timeout = call_args[0][1]
                     assert timeout >= 0, "socket timeout must be nonnegative"
 
+    def test_ssl_server_hostname(self):
```
StephenSorriaux, line 33: If possible, I probably would have liked a test around the new `check_hostname` param. What do you think?

kazoo/handlers/utils.py
```diff
@@ -241,7 +243,7 @@ def create_tcp_connection(
...
             # elect to use it.
-            context.check_hostname = False
+            context.check_hostname = check_hostname
             context.verify_mode = (
                 ssl.CERT_REQUIRED if verify_certs else ssl.CERT_NONE
```
StephenSorriaux, line 248: Based on the comment L242, it seems we might need to raise a `ValueError` or something if `check_hostname=True` and `verify_certs=False`. Maybe in the client class when checking the given params. What do you think?

StephenSorriaux (commented):
Thank you for this PR. Just have some comments here and there. No worries about the segfault, the pypy testing pipeline just triggers it from time to time, it probably relates to the GA infrastructure, but I really never investigated it since I was not able to replicate it locally.

kazoo/client.py
jessesightler-redhat: I have added a line to the description.

kazoo/tests/test_utils.py
jessesightler-redhat: Good idea, test added

kazoo/handlers/utils.py
jessesightler-redhat: Done :)

kazoo/handlers/utils.py
jessesightler-redhat: Agreed. I've added an error for this case.

StephenSorriaux (dismissed):
Thanks a lot, that's great.

I will take care of doing a squash of the commits when merging, so that the final commit message follows our guidelines.

I will also wait a little for another maintainer to gives another feedback.

kazoo/handlers/utils.py
```diff
@@ -237,11 +239,14 @@ def create_tcp_connection(
 
             # Load default CA certs
             context.load_default_certs(ssl.Purpose.SERVER_AUTH)
+            if check_hostname and not verify_certs:
+                raise ValueError(
+                    "Error, if check_hostname is specified "
+                    + "verify_certs must be False"
```
jeffwidman, line 245: `"Error"` is redundant, and we could probably word this a little more concisely:

```suggestion
                    "verify_certs must be False when check_hostname is specified"
```

Feel free to break across lines if the linter complains...

kazoo/handlers/utils.py
```diff
@@ -237,11 +239,14 @@ def create_tcp_connection(
...
+                )
             # We must set check_hostname to False prior to setting
             # verify_mode to CERT_NONE.
-            # TODO: Make hostname verification configurable as some users may
-            # elect to use it.
```
jeffwidman, line 243: ❤️ for solving this!

jeffwidman (approved):
I didn't review it deeply as I'm not deeply familiar with SNI, but what I saw seems reasonable. 

Thanks for plumbing it through and also including tests.

kazoo/handlers/utils.py
jessesightler-redhat: Thanks... I've incorporated your suggestion for the error text as well.

kazoo/handlers/utils.py
```diff
@@ -241,7 +241,8 @@ def create_tcp_connection(
             context.load_default_certs(ssl.Purpose.SERVER_AUTH)
             if check_hostname and not verify_certs:
                 raise ValueError(
-                    "verify_certs must be False when check_hostname is specified"
+                    "verify_certs must be False when"
```
ceache, line 244: From the check on line 242, think the intent is to make sure that the hostname validation is *in addition* to the cert validation, correct?
If so, I think the error msg is misleading.
How about something like "verify_certs must also be True when check_hostname is True" ?

jessesightler-redhat: Agreed. I have tweaked the wording a bit now.

ceache (approved):
LGTM. Thanks a lot for your contribution!

ceache:
@StephenSorriaux I believe you wanted to handle the squash & merge?

StephenSorriaux (approved):

PR #763, opened 2025-08-27
removing tiny remaining bits of Python2 hybridation

a-detiste:
Hi,

I noticed these tiny glitches while checking new Debian upload of this package.

Have a nice day


jeffwidman (approved):
Thank you!

PR #776, opened 2026-04-25
feat(testing): updated ensure-zookeeper-env.sh to match minimum version of zk

ThosRTanner:
These changes allow people to build locally with a reasonable chance of success making PRs a bit less likely to fail.

Also updated `.gitignore` to cover a few local files when developing with visual studio

## Why is this needed?

* Building kazoo locally would fail because the `ZOOKEEPER_VERSION` env variable isn't exported 
* The currently selected default version was way short of the minimum supported version, resulting in odd failures when building locally, and odd failures in PR builds
* Also added some stuff to `.gitignore` to assist in local builds with visual studio code

## Proposed Changes

 - Move all optional settings into ensure-zookeeper-env.sh out of `workflows/testing.yml`
 - Change `ensure-zookeeper-env.sh` so if `ZOOKEEPER_VERSION` wasn't supplied, to select the lowest (at the time of writing) supported version of zookeeper and export the variable
 - Updated `.gitignore` to include `.*_cache` and `venv*` as this makes them greyed out in visual studio code

No


StephenSorriaux (dismissed):
Thank you for taking care of my old TODO.

ThosRTanner:
> Thank you for taking care of my old TODO.
No probs. I think you need to re-review it after I merged it but at least it's not complaining about coverage now!

StephenSorriaux (dismissed):

StephenSorriaux (dismissed):

StephenSorriaux:
Can you please amend your commit and match the contribution guideline? Just adding `feat(testing):` before the current message would be enough. That helps when we auto generate the changelog for a release.

ThosRTanner:
@StephenSorriaux oops - hope that worked. it dismissed your review again...

StephenSorriaux (approved):
Thanks a lot!

PR #784, opened 2026-05-08
chore(testing): Pin openssl version to latest version providing crypto module

ThosRTanner:
## Why is this needed?

The latest version of openssl has dropped the crypto module and so nothing builds

## Proposed Changes

Adding a pin of pyOpenSSL to the previous to latest version

No


StephenSorriaux (approved):
Thank you.

PR #797, opened 2026-09-01
typing: propagate result types through async APIs

bringhurst:
## Why is this needed?

Kazoo's async methods return a bare `IAsyncResult`, so `get()` becomes `Any` and synchronous wrappers need casts. This makes the public annotations less useful.

This is the first step toward typing the full request path. A follow-up can tie each wire request to its response with a generic `_submit` helper, then carry that type through the pending queue.

## Proposed Changes

  - Make `IAsyncResult` generic in the value it carries.
  - Add result types to the public async client methods.
  - Remove casts that are no longer needed.
  - Type ignored callback and handler-specific `wait()` returns as object.
  - Add static checks for the public API.

No. Runtime APIs and behavior stay the same.

### DIFF
--- a/kazoo/client.py
+++ b/kazoo/client.py
@@ -59,6 +59,7 @@ from kazoo.protocol.serialization import (
     SetData,
     Sync,
     Transaction,
+    Transaction_Response,
 )
 from kazoo.protocol.states import (
     Callback,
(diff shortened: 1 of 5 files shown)


ThosRTanner:
That's nice! Takes something off my todo list as well!

kazoo/interfaces.py
```diff
@@ -158,6 +160,7 @@ def join(self, timeout: float | None = None) -> None:
 
 
 SpawnedFunc = Callable[..., None]
+_ResultT = TypeVar("_ResultT", default=Any)
```
StephenSorriaux, line 163: we did not pin typing-extensions and it seems this use of TypeVar requires typing-extensions >= 4.4.0, we should probably pin it

kazoo/client.py
StephenSorriaux, line 1392: Not really related to this PR but that might need a change? `result.get()` is not a bool in case of `create_async()`

kazoo/interfaces.py
StephenSorriaux: handled in https://github.com/python-zk/kazoo/pull/798

Issue #790, opened 2026-05-15
Consider deprecating python3.8


ThosRTanner:
It was end-of-lifed 2 years ago

https://devguide.python.org/versions/

PR #798, opened 2026-09-04
chore: merge constraints.txt and setup.cfg into pyproject.toml to simplify setup

### LINKED ISSUE #786: Move to using pyproject.toml and requirements.txt rather than setup.cfg+pyproject.toml+constraints.txt
The current system is confusing and requires constraints to be updated in multiple places. Moreover it renders the python caching a bit pointless as it's based off requirements.txt and - there isn't one.

Furthermore, the setup-python has its own support for pip caching and we should use that in preference to the explicit action-cache currently being used

ThosRTanner:
Fixes #786 

## Why is this needed?

Currently constraints are specified in 3 places which makes it hard to track what you need to change. I'm not an out and out lover of `pyproject.toml` but having all module setup in one place makes it easier to control package version dependencies.

All that is left in setup.cfg is the `tag_build = dev` to tag dev builds as such, as I've been unable to find an obvious equivalent in `pyproject.toml`

## Proposed Changes

  - Move `constraints.txt` and `setup.cfg` info into `pyproject.toml`. This also requires us to:
    - Update setuptools to a version that supports dependency groups
    - Update the pytest version for python3.8 to the latest version that still supports python3.8
    - Update the required version of tox
  - Differentiate between the project's optional dependencies and the internal tooling dependencies
  - Update `tox.ini` to use new dependency groups
  - Remove the top level `Makefile` which has not been maintained  and refers to non existent requirements file
  - Move control of sphinx dependencies into `docs/requirements.txt`
     - also make `tox -e docs` actually generate the docs.
   - Prune `MANIFEST.in` as a lot of it has no effect

The internal dependencies are no longer leaked from `setup.cfg`, and the top level `Makefile` has been removed. This shouldn't affect anyone, but one never knows...

## Codecov report

This appears to be because it's two commits out of date. This change hasn't affected what code is covered during tests


.readthedocs.yaml
```diff
@@ -26,7 +26,4 @@ formats: []
 # Optionally declare the Python requirements required to build your docs
 python:
   install:
-  - method: pip
-    path: .
-    extra_requirements:
-    - docs
+  - requirements: docs/requirements.txt
```
StephenSorriaux, line 29: Will this be enough? Before it seems to do `pip install . [docs]`, so kazoo + docs subgroup, now it seems to only install the docs libs, without kazoo? I think it might be missing the `typing-extensions`

Makefile
StephenSorriaux, line 1: Reference should be removed from `MANIFEST.in` too

pyproject.toml
```diff
@@ -1,9 +1,98 @@
...
+    "Programming Language :: Python :: 3.12",
+    "Programming Language :: Python :: 3.13",
+    "Programming Language :: Python :: 3.14",
+    "Programming Language :: Python :: Implementation :: CPython",
+    "Programming Language :: Python :: Implementation :: PyPy",
+    "Topic :: Communications",
+    "Topic :: System :: Distributed Computing",
+    "Topic :: System :: Networking",
+]
+dependencies = ["typing-extensions"]
```
StephenSorriaux: I mentionned it here https://github.com/python-zk/kazoo/pull/797#discussion_r3941304734, I think it needs to be pinned, probably `>=4.5` because of `deprecated` but not 100% sure

pyproject.toml
```diff
@@ -1,9 +1,98 @@
...
+    "coverage==6.3.2; python_version == '3.8'",
+    "coverage==7.10.7; python_version > '3.8'",
+    {include-group = "pytest"},
+    "pytest-cov==3.0.0; python_version == '3.8'",
+    "pytest-cov==7.0.0; python_version > '3.8'",
+    "pytest-timeout==2.2.0; python_version == '3.8'",
+    "pytest-timeout==2.4.0; python_version > '3.8'",
+    {include-group = "test-extras"},
+]
+mypy = [
```
StephenSorriaux, line 74: May I ask why it was not added as a `project.optional-dependencies` too (`[typing]`)?

pyproject.toml
```diff
@@ -1,9 +1,98 @@
...
+    "types-mock",
+    "types-objgraph",
+    "types-pyjks",
+]
+
+[tool.setuptools]
+zip-safe = false
+include-package-data = true
+packages = {find = {}}
+license-files = ["LICENSE"]
```
StephenSorriaux, line 88: Would it be better to put it in the `[project]` section instead? https://packaging.python.org/en/latest/specifications/pyproject-toml/#license-files

pyproject.toml
```diff
@@ -1,9 +1,98 @@
...
+    "Programming Language :: Python :: 3.14",
+    "Programming Language :: Python :: Implementation :: CPython",
+    "Programming Language :: Python :: Implementation :: PyPy",
+    "Topic :: Communications",
+    "Topic :: System :: Distributed Computing",
+    "Topic :: System :: Networking",
+]
+dependencies = ["typing-extensions"]
+
+[project.urls]
```
StephenSorriaux, line 39: May I ask for `Homepage = "https://kazoo.readthedocs.io"` to be added?

tox.ini
```diff
@@ -41,27 +41,27 @@ commands =
...
 usedevelop = True
 commands = flake8 {posargs} {toxinidir}/kazoo
 
 [testenv:black]
-extras =
-deps =
-    black
+dependency_groups = black
 usedevelop = True
 commands = black --check {posargs: {toxinidir}/kazoo {toxinidir}/kazoo}
```
StephenSorriaux: Seems like we had 2 times the same arg here? weird

Makefile
ThosRTanner: Oops. Done.

Actually, I gave `MANIFEST.in` a good going over and reduced it to minimum size and checked the before and after contents matched.

pyproject.toml
ThosRTanner: Ah, looks like I have 4.13.2. I'll pin it to at least that.

pyproject.toml
ThosRTanner: typing? or typing-extensions. You need typing-extensions because the some of the types defined in there are used in places where the interpreter can't skip over it, sadly. The whole of mypy is a dependency group so it's only installed if you're explicitly installing it - which is what tox does.

tox.ini
ThosRTanner: yeah, that is definitely weird.

I removed the 2nd.

.readthedocs.yaml
ThosRTanner: Um, yes, not it wasn't. I went and destroyed my tox environment and rebuilt it and found a couple of dependencies missing.

pyproject.toml
ThosRTanner: not supported for python 3.8 sadly.

pyproject.toml
ThosRTanner: The same url as the "Documentation" page? Or replace the documentation page?

I've assumed the first but let me know if you want it changed.

pyproject.toml
StephenSorriaux: Yes, `typing`. It was in `options.extras_require` in `setup.cfg` and included all deps linked to typing. My understanding was that an user could `pip install kazoo[typing]` to install it if they wanted to. Now, it does not seem possible anymore? Is it what we want?

pyproject.toml
StephenSorriaux: Yes, the same.

.readthedocs.yaml
StephenSorriaux: Thanks, is the change OK now? I don't see anything new but I might be missing something.

Makefile
StephenSorriaux: Thank you.

pyproject.toml
ThosRTanner: Done then

.readthedocs.yaml
ThosRTanner: yes, it's fine now. it (tox) creates a new environment and installs the necessary dependencies in there from docs/requirements.txt

pyproject.toml
ThosRTanner: ah, ok. I hadn't realised that.

However, doing what is suggested in the literature doesn't work for python3.8 and mypy fails completely.

I've added a dependency group `typing` and for python3.9+ (well, the first pip that was released after python3.8 support was dropped) you can use `pip install --group typing`. Would that be a reasonable approach?

StephenSorriaux: Sounds very reasonable to me, thank you.

StephenSorriaux (approved):

ceache:
Do we really want to keep supporting 3.8?
I appreciate the effort you put in and all the workarounds you had to figure out, but 3.8 EoL'd in 2024, 3.9 EoL'd October 2025, 3.10 will EoL in a month or so. No bug fix, no security patches.
I don't think we have to keep supporting so many dead snakes. Maybe the "last dead" one, 3.9 today, but why do any more than that?

StephenSorriaux:
I think that makes sense this PR includes Python 3.8 since it currently what the project supports. Changes can be made in another PR to drop Python 3.8 (and take care of the current Python 3.8 workarounds) if that's the way we want to go.

ceache (approved):

ThosRTanner:
@ceache @StephenSorriaux I already raised a ticket to drop 3.8 support - #790

Issue #594, opened 2020-03-10
SASL reconnection error


jeblair:
When using SASL authentication, if a kazoo client is disconnected, all future API calls will return NoAuthError on reconnect.

The issue appears to be that the connection object maintains the state of the SASL request in the .sasl_cli attribute, but that is not reset on disconnection.  That means that when the reconnection attempt is made, it thinks that the SASL request is already completed, so it is not performed.

I am able to temporarily work around this in my app by setting `client._connection.sasl_cli = None`.  Perhaps something similar in the connection handling code would address the problem.

ceache:
Hi,

Thank you for your report.

Can you clarify a little bit what state transitions your client is going
through when you say "disconnected"?

A SUSPENDED connection as the client transitions to a new server in the
ensemble?
A LOST connection that you subsequently reconnect?
A stop()/start()?

Thanks,

PR #806, opened 2026-09-13
fix(protocol): propagate ConnectionDropped during SASL handshake to enable failover retry

ceache:
Addresses #594

### Why is this needed?
When a ZooKeeper ensemble node drops the connection during SASL negotiation (e.g. server node restarting or leader election occurring during reconnect), `_authenticate_with_sasl` in `kazoo/protocol/connection.py` caught `ConnectionDropped` and re-raised it as `AuthFailedError("Connection dropped in SASL")`.

In `_connect_attempt`, catching `AuthFailedError` treated this as an unrecoverable authentication credential failure and returned `STOP_CONNECTING`. This permanently halted the connection loop instead of retrying other available ensemble hosts.

### Changes
- Raise `ConnectionDropped` instead of `AuthFailedError` when `_read_header` encounters `ConnectionDropped` during SASL handshake, allowing the connect retry loop to fail over to remaining hosts.


StephenSorriaux (dismissed):
Great catch, thank you.

StephenSorriaux (approved):

Issue #775, opened 2026-04-24
The continuous integration is a bit flaky and prone to timeouts and other issues


ThosRTanner:
I can do a PR and one (or more) integration tests will fail randomly with timeout connecting to the server. I can then do an empty commit and the integration tests all run fine.

PR #807, opened 2026-09-13
test(integ): sync follower in test_connect_auth to avoid replication race

ceache:
Addresses #775

### Why is this needed?
In `test_connect_auth` (`kazoo/tests/test_client.py`), a client creates an ACL-protected node `/1` followed by a fixed `time.sleep(0.1)` before another client attempts to read it.
In ZooKeeper, follower reads are sequentially consistent, but may briefly lag behind the leader until committed transactions are applied locally. If the test client connects to a lagging follower node in the cluster, `get("/1")` intermittently raises `NoNodeError` instead of `NoAuthError`, causing CI test flakiness.

### Changes
- Replace `time.sleep(0.1)` with `self.client.sync("/1")` to explicitly flush the replication channel from the leader to the connected follower before issuing the read.


StephenSorriaux (dismissed):
Thank you.

StephenSorriaux (approved):

PR #810, opened 2026-09-13
fix(protocol): handle SASLException and SessionClosedRequireSaslError in connection attempt

ceache:
Addresses #594

### Why is this needed?
During connection attempts, if SASL authentication fails with a `SASLException` (e.g. from `puresasl` library errors or missing SASL support), or if the ZooKeeper server closes the session requiring SASL authentication (`SessionClosedRequireSaslError` / code -124), these exceptions were unhandled in `ConnectionHandler._connect_attempt`.

Because `_connect_attempt` only caught `AuthFailedError`, these exceptions either caused `zk_loop` to fail with an unhandled exception or caused retry loops rather than stopping and cleanly notifying the client that authentication failed.

### Changes
- Catch `SASLException` alongside `AuthFailedError` in `ConnectionHandler._connect_attempt`, reset retry counter, notify `client._session_callback(KeeperState.AUTH_FAILED)`, and return `STOP_CONNECTING`.
- Catch `SessionClosedRequireSaslError` in `ConnectionHandler._connect_attempt`, reset retry counter, log informative warning, notify `client._session_callback(KeeperState.AUTH_FAILED)`, and return `STOP_CONNECTING`.
- Add unit tests in `TestConnectionAuthExceptions` in `test_connection.py`.
- Update `CHANGES.md` under Unreleased.


StephenSorriaux (approved):
Thank you.

PR #811, opened 2026-09-13
Chore: Remove outdated zookeeper version checks

### LINKED ISSUE #787: There are still checks for zookeeper versions less than minimum supported
Given we only support 3.6 and above, there's little point in checking in the tests that the zookeper version is 3.4 or above (that's not the only one).

We should check that at the start of testing and be done with it there.

ThosRTanner:
Fixes #787

## Why is this needed?

Well, mostly because it's confusing to read the code. We have checks for zookeeper version 3.4 and 3.5, neither of which are supported.

## Proposed Changes

  - Ensure zookeeper version is a permitted version when setting up the testing environment in tox.
  - Remove all checks for zookeeper version
  - Remove `CI` and `CI_*` environment variable handling (in particular tox renames the CI environment variable so it doesn't really help checking it). There is a hardcode delay which will now be applied all the time though possibly we should use something a little more programmatic.

No

## Codecov report

Probably because codecov is checking from10 commits behind main, possibly related to https://github.com/codecov/codecov-action/issues/1855


ensure-zookeeper-env.sh
```diff
@@ -2,9 +2,33 @@
...
+HERE=$(pwd)
 ZOO_BASE_DIR="$HERE/zookeeper"
-export ZOOKEEPER_VERSION=${ZOOKEEPER_VERSION:-3.6.4}
+
+# Supported Zookeeper versions. NB Must match the versions in the build matrix in 
+# .github/workflows/testing.yml
+ZOOKEEPER_VERSIONS=(3.6.4 3.7.2 3.8.3 3.9.1)
+
+if [ -z "$ZOOKEEPER_VERSION" ]; then
+    echo "No ZOOKEEPER_VERSION specified, using default: ${ZOOKEEPER_VERSIONS[0]}"
```
StephenSorriaux: i would probably prefer the default one to be the latest one

StephenSorriaux (dismissed):
Thank you.

ensure-zookeeper-env.sh
ThosRTanner: Done

StephenSorriaux (approved):

ThosRTanner:
@StephenSorriaux the codecov report is hysterical "no files covered by tests were changed - coverage down by 0.08% - failed"

ThosRTanner:
github it doesn't seem to like auto merges much if you get out of step with the base branch :-(

PR #821, opened 2026-09-16
chore: use setup-python to cache dependencies

ThosRTanner:
## Why is this needed?

The existing `cache` setup is gnarly and depended on `requirements.txt` (which didn't exist).

## Proposed Changes

  - Use `setup-python` github action to control pip caches rather than separate `cache` github action.

No


StephenSorriaux (dismissed):
Thank you.

StephenSorriaux (approved):

PR #822, opened 2026-09-17
chore: Fix CI zookeeper caching.

### LINKED ISSUE #785: PR tests occasionally fail due to an error downloading zookeeper from apache
Can't this be cached? See https://github.com/actions/cache/tree/main

ThosRTanner:
Fixes #785

## Why is this needed?

zookeeper versions aren't being retrieved from the cache, causing repeated downloads,
## Proposed Changes

- Add the zookeeper version to the cache key (and as it's java, make it cross platform to save a bit of space).
- Change the python version used for windows tox to 3.12 to match the version in python-setup (similarly to how the main test workflow works)

No


.github/workflows/testing.yml
```diff
@@ -119,12 +118,16 @@ jobs:
     name: Windows - Sanity test using a single version of Python and ZK
 
     runs-on: windows-latest
+    env:
+      ZOOKEEPER_VERSION: "3.9.1"
+      PYTHON_VERSION: "3.10"
```
StephenSorriaux: The initial intent was for Windows to be tested with python 3.12, not 3.10. That's my bad when I bumped the versions and forgot the tox env.

ThosRTanner: updated accordingly

PR #824, opened 2026-09-19
chore(ci): configure Codecov build counts and status thresholds

ceache:
## Summary

Adds a `.codecov.yml` configuration file to configure Codecov build count expectations, realistic patch coverage targets, and informational status reporting, and adds the Codecov status badge to `README.md`.

### Problem
Without a `.codecov.yml` file, Codecov applies its default zero-tolerance policies:
1. **`codecov/patch`**: Requires strict 100% coverage on every single modified line in a PR, failing PR checks whenever a new branch, fallback path, or defensive exception check is not directly exercised.
2. **`codecov/project`**: Compares PR coverage against `master` with a 0% drop threshold, causing PRs to fail over minor fractional fluctuations (e.g. -0.05%).
3. **Upload count mismatch**: Codecov expects the same number of report uploads as the base commit and flags discrepancies when matrix dimensions vary.

### Changes
- **`after_n_builds: 21`**: Tells Codecov to wait for all 21 test matrix jobs (7 Python versions × 3 ZooKeeper versions) before evaluating coverage and comparing against base. Includes an inline comment noting that this value should be updated if matrix dimensions change.
- **`coverage.status.project.default`**:
  - `target: auto`
  - `threshold: 1%`: Accommodates normal fractional fluctuations without flagging failure.
  - `informational: true`: Ensures coverage reports status informatively without blocking PR CI.
- **`coverage.status.patch.default`**:
  - `target: 80%` & `threshold: 5%`: Sets a realistic patch coverage expectation.
  - `informational: true`: Avoids hard CI failures for defensive error handling or unreachable branches.
- **`README.md`**: Added the official Codecov coverage badge for `python-zk/kazoo:master`.


ThosRTanner:
Would it be a good idea to add a comment to testing.yml saying if the matrix changes then you need to update .codecov.yml? Otherwise it's more likely to be missed.

ceache:
> Would it be a good idea to add a comment to testing.yml saying if the matrix changes then you need to update .codecov.yml? Otherwise it's more likely to be missed.

I think it makes sense to have both these files referring to each other, yes. Let me do that real quick

